PR #2, opened 2016-04-12
[JENKINS-30206] Test demonstrating that new branch properties are invoked

FrantaM:
[JENKINS-30206](https://issues.jenkins-ci.org/browse/JENKINS-30206)

The problem as I see it is as follows: 

- Branch [is compared](https://github.com/jenkinsci/branch-api-plugin/blob/48717c40c8b0f778073a595a0087f39274d71238/src/main/java/jenkins/branch/Branch.java#L167) by source id and head
- Workflow factory [will not update Branch](https://github.com/jenkinsci/workflow-multibranch-plugin/blob/0d311cdc459539b219927ac9124054d18b986385/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactory.java#L62) in Workflow job when the current one is equal to [the new one](https://github.com/jenkinsci/branch-api-plugin/blob/48717c40c8b0f778073a595a0087f39274d71238/src/main/java/jenkins/branch/MultiBranchProject.java#L259)

As you can see the branch source has changed even though source's id and branch head have not. Note that this affect not only branch properties but also, for example, checkout credentials.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProjectTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProjectTest.java
@@ -27,6 +27,8 @@ package org.jenkinsci.plugins.workflow.multibranch;
 import hudson.ExtensionList;
 import hudson.model.DescriptorVisibilityFilter;
 import hudson.model.Item;
+import hudson.model.Job;
+import hudson.model.Run;
 import hudson.model.listeners.ItemListener;
 import hudson.plugins.git.GitSCM;
 import hudson.scm.ChangeLogParser;
@@ -42,6 +44,7 @@ import jenkins.branch.BranchProperty;
 import jenkins.branch.BranchPropertyDescriptor;
 import jenkins.branch.BranchSource;
 import jenkins.branch.DefaultBranchPropertyStrategy;
+import jenkins.branch.JobDecorator;
 import jenkins.plugins.git.GitSCMSource;
 import jenkins.scm.api.SCMHead;
 import jenkins.scm.api.SCMSource;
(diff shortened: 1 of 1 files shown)


src/test/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProjectTest.java
```diff
@@ -191,4 +194,36 @@ public DescriptorImpl() {
         }
     }
 
+    @Issue("JENKINS-30206")
```
jglick: If this is something you claim is currently failing, then you should add `@Ignore` so that the PR itself is mergeable.

src/test/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProjectTest.java
```diff
@@ -191,4 +194,36 @@ public DescriptorImpl() {
         }
     }
 
+    @Ignore
+    @Issue("JENKINS-30206")
```
jglick: Could you put this in `JobPropertyStepTest` please, since I presume this is intended as a follow-up to https://github.com/jenkinsci/workflow-plugin/pull/199?

jglick:
As noted in the JIRA issue, whatever bug there is does not really affect multibranch Pipeline jobs, since (as far as I am aware) there are no `BranchProperty`s applicable to this multibranch type. So what bug _symptom_ are you seeing?

jglick:
> this affect not only branch properties but also, for example, checkout credentials

If that is the problem that you care about, then JENKINS-30206 does not really describe the bug, and this test does not reproduce it.

FrantaM:
> So what bug _symptom_ are you seeing?

As described in [JENKINS-30206](https://issues.jenkins-ci.org/browse/JENKINS-30206?focusedCommentId=235736&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-235736): when you add a new **branch** property or edit an existing one it is not reflected in existing workflow jobs.

> (as far as I am aware) there are no `BranchProperty`s applicable to this multibranch type

Anyone can write one, right?

src/test/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProjectTest.java
FrantaM: It is not as this has nothing to do with job properties or parameter definitions.

jglick:
> when you add a new branch property

There are none to add. The _Add_ button is not even displayed for multibranch Pipeline projects.

> Anyone can write one, right?

They could, but I do not advise it; I went out of my way to suppress them all.

jglick:
Better to have this pull in https://github.com/jenkinsci/branch-api-plugin/pull/43 (1.9-beta-1) and do a real test.

jglick:
#12 adds a more realistic test (and fixes the bug), so I am closing this. Thanks though!

PR #3, opened 2016-04-12
Warn if jenkinsfile is encountered when searching for Jenkinsfile

jglick:
@recena claimed to have had some trouble diagnosing this, so make it more apparent.

Actually accepting a branch with `jenkinsfile` is more complicated, due to usage from `SCMVar`, and this does not try—just alerts the user to the mistake.

The downside is more time (and perhaps network traffic) spent rejecting non-multibranch-ready branches and repositories. And in the case of `github-branch-source`, extra noise as the source prints messages about `jenkinsfile` not existing, which is probably more confusing than helpful because it implies that `jenkinsfile` _would_ be accepted. So we need some refined solution TBD.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProject.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProject.java
@@ -31,6 +31,7 @@ import hudson.model.TaskListener;
 import hudson.model.TopLevelItem;
 import hudson.scm.SCMDescriptor;
 import java.io.IOException;
+import java.util.Locale;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import jenkins.branch.BranchProjectFactory;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProject.java
```diff
@@ -50,9 +51,17 @@
...
     static final String SCRIPT = "Jenkinsfile";
+    private static final String SCRIPT_LOWERCASE = SCRIPT.toLowerCase(Locale.ENGLISH);
     static final SCMSourceCriteria CRITERIA = new SCMSourceCriteria() {
         @Override public boolean isHead(SCMSourceCriteria.Probe probe, TaskListener listener) throws IOException {
-            return probe.exists(SCRIPT);
+            if (probe.exists(SCRIPT)) {
+                return true;
+            } else {
+                if (probe.exists(SCRIPT_LOWERCASE)) {
+                    listener.getLogger().println("Found " + SCRIPT_LOWERCASE + " rather than " + SCRIPT + ", skipping");
```
recena: Let me try the indentation on the listener...

recena: @jglick I prefer to keep the original behavior due to this `listener` breaks the log message structure that we are using in GitHub Branch Source.

jglick: Yeah I mentioned that in the PR description.

jglick:
If https://github.com/jenkinsci/github-branch-source-plugin/issues/50 supersedes this, then fine to close this.

jglick:
(Though we would need similar refinements to other branch sources.)

PR #4, opened 2016-04-14
[JENKINS-34235] Wrong plugin short-name in the URL pattern

recena:
[JENKINS-34235](https://issues.jenkins-ci.org/browse/JENKINS-34235)

@reviewbybees especially @amuniz

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProject.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProject.java
@@ -79,7 +79,7 @@ public class WorkflowMultiBranchProject extends MultiBranchProject<WorkflowJob,W
         }
 
         public String getIconFilePathPattern() {
-            return "plugin/workflow-images/:size/pipelinemultibranchproject.png";
+            return "plugin/workflow-multibranch/:size/pipelinemultibranchproject.png";
         }
 
         @Override public TopLevelItem newInstance(ItemGroup parent, String name) {


src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProject.java
```diff
@@ -79,7 +79,7 @@ public String getDescription() {
         }
 
         public String getIconFilePathPattern() {
-            return "plugin/workflow-images/:size/pipelinemultibranchproject.png";
+            return "plugin/workflow-multibranch/:size/pipelinemultibranchproject.png";
```
amuniz: Why was it `workflow-images` before?? Did that even exist at some point?

amuniz:
:bee:

src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProject.java
recena: It was my wrong, maybe a copy&paste??? :panda_face:

recena: In the original PR was fine https://github.com/jenkinsci/pipeline-plugin/pull/376/files#diff-fc4cf5ce663b563407ec5283d4866c34R82

I found the problem. I had to do a replace of `multibranch` to adapt the patch file :sob:

jtnord:
:bee:

PR #6, opened 2016-04-19
[JENKINS-34200] Redirect to computation/console after creating a Multibranch project

recena:
[JENKINS-34200](https://issues.jenkins-ci.org/browse/JENKINS-34200)

downstream of https://github.com/jenkinsci/cloudbees-folder-plugin/pull/58

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -89,7 +89,7 @@ THE SOFTWARE.
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>cloudbees-folder</artifactId>
-            <version>5.8</version>
+            <version>5.9-SNAPSHOT</version>
         </dependency>
         <dependency>
             <groupId>${project.groupId}</groupId>


amuniz:
:bee:

oleg-nenashev:
:bee:

jglick:
You forgot to `mvn deploy` upstream.

Anyway :bug:, this should not be here, rather in `MultiBranchProject` + `OrganizationFolder` (or perhaps even more broadly in `ComputedFolder`, meaning there is no need for a public API at all).

src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProject.java
```diff
@@ -64,6 +64,10 @@ public WorkflowMultiBranchProject(ItemGroup parent, String name) {
         return new WorkflowBranchProjectFactory();
     }
 
+    @Override protected String getSuccessfulDestination() {
+        return "computation/console";
```
jglick: Should be using `FolderComputation.getSearchUrl` as the base, in this case resulting in `"indexing/console"`.

recena:
@jglick Well, again different PoVs about a simple design.

recena:
@jglick I didn't forget, it was underway

jglick:
> I didn't forget, it was underway

Best to `deploy` _before_ filing the downstream PR to avoid an initial build error.

recena:
@jglick My first intention was to work on `ComputedFolder` but it is not possible to define a redirection:

```
@RequirePOST
@Override
public void doConfigSubmit(StaplerRequest req, StaplerResponse rsp) throws IOException, ServletException, Descriptor.FormException {
    super.doConfigSubmit(req, rsp);
    scheduleBuild();
}
```

Obviously without duplicate code from `doConfigSubmit`.

jglick:
> different PoVs

Well, my point of view is that it is incorrect to be doing this customization here, where it will fail to apply to other multibranch project types, which is an artificial limitation. It will also fail to apply to organization folders.

jglick:
> My first intention was to work on ComputedFolder but it is not possible to define a redirection

If you want to apply the change to any `ComputedFolder`, just take the current upstream patch, make it `@Restricted(NoExternalUse.class)` (at least unless and until another use case is identified), and override in `ComputedFolder` (calling `getSearchUrl`).

recena:
@jglick 

> where it will fail to apply to other multibranch project types

Why? Other implementations of `MultiBranchProject` will be able to override the method `getSuccessfulDestination` or simply do nothing and get the current behavior.

recena:
@jglick It seems you are telling two different alternatives. Do you have any preference about one of them?

recena:
@jglick 

> If you want to apply the change to any `ComputedFolder`, just take the current upstream patch, make it `@Restricted(NoExternalUse.class)` (at least unless and until another use case is identified), and override in `ComputedFolder` (calling `getSearchUrl`).

I wanted to apply these changes in `ComputedFolder` but with a public API in order to allow that its implementations can decided what do.

I disagree with _All `ComputedFolder` implementations_ will be redirected to `computation/console`.

recena:
@jglick To me you have used the bug to block this PR because you prefer other design.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProject.java
recena: @jglick I did it [here](https://github.com/jenkinsci/cloudbees-folder-plugin/pull/58/files#diff-73ee53b9a4352457ad14440d3f6692c0R242) but it is strange that:

1. if you click on _Branch Indexing_ the url contains `computation/` but...
1. if you add a new Multibranch project, you are redirected to `indexing/` blah blah

IMO, the URLs in the left menu (actions) are not valid currently.

jglick:
> It seems you are telling two different alternatives. Do you have any preference about one of them?

Modest preference for a fix that applies to any `ComputedFolder`, since

* it is simpler—no API change or multi-plugin patch required
* in all use cases for `ComputedFolder` that I can think of, the same behavioral change is desired

> with a public API in order to allow that its implementations can decided what do

You should not introduce an API without preparing at least one use case for it.

> I disagree with _All `ComputedFolder` implementations_ will be redirected to `computation/console`.

OK; why do you disagree? Which current or plausibly proposed implementation do you think would better redirect to `.`?

jglick:
> if you click on _Branch Indexing_ the url contains `computation/`

Sounds like a bug somewhere else.

recena:
@jglick 

> Sounds like a bug somewhere else

I'll file a Jira issue for that.

recena:
@jglick 

> Sounds like a bug somewhere else

https://github.com/jenkinsci/cloudbees-folder-plugin/pull/60

jglick:
Is this now obsolete?

recena:
@jglick Not to me. I think if we upgrade the dependency the UX using `workflow-multibrach` will be better.

jglick:
> if we upgrade the dependency the UX […] will be better

Such a dependency upgrade does nothing for users except force them to upgrade when they may have had a reason to hold back. Anyone installing the plugins anew, or routinely accepting plugin updates, would get the newer dependency anyway. So :bug: on an unnecessary change.

recena:
I disagree but you decide.

recena:
You are considering the dependencies as the way to declare what you need in compilation time, in other words, _what you need from their APIs_.

My PoV is _what you need to get the full feature list_ in a declarative way thought dependencies. A full user experience using the plugin.

Anyway, you are the maintainer, you decide.

PR #8, opened 2016-05-11
[JENKINS-33819] New API to create WorkflowMultiBranch projects specifying an OrphanedItemStrategy

recena:
[JENKINS-33819](https://issues.jenkins-ci.org/browse/JENKINS-33819)

downstream of https://github.com/jenkinsci/cloudbees-folder-plugin/pull/61
downstream of https://github.com/jenkinsci/branch-api-plugin/pull/41

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -84,12 +84,12 @@ THE SOFTWARE.
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>branch-api</artifactId>
-            <version>1.1</version>
+            <version>1.8-SNAPSHOT</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>cloudbees-folder</artifactId>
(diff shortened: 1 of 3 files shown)


recena:
Deploying spanshots...

jglick:
As noted in the upstream PRs, using a setter would obviate the need for this PR to even exist.

amuniz:
> As noted in the upstream PRs, using a setter would obviate the need for this PR to even exist.

Why? I think this is still needed. The setter should be called here and not in `OrganizationFolder`, so each `MultiBranchProjectFactory` implementation can decide if inherit the configuration or not.

jglick:
> so each `MultiBranchProjectFactory` implementation can decide if inherit the configuration or not

I do not see any particular reason for a factory to be involved in this decision. If such a reason is found in the future, an API addition could be made for it.

recena:
I agreed with the proposal in this PR, for that reason I sent the PR 😄

PR #10, opened 2016-05-24
[JENKINS-34596] Added readTrusted step

jglick:
[JENKINS-34596](https://issues.jenkins-ci.org/browse/JENKINS-34596)

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/multibranch/ReadTrustedStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/multibranch/ReadTrustedStep.java
@@ -1,0 +1,188 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2016 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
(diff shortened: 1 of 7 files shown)


src/main/java/org/jenkinsci/plugins/workflow/multibranch/ReadTrustedStep.java
```diff
@@ -0,0 +1,188 @@
...
+                    throw new AbortException("Could not determine exact tip revision of " + branch.getName());
+                }
+                build.addAction(new SCMRevisionAction(tip));
+            }
+            SCMRevision trusted = scmSource.getTrustedRevision(tip, listener);
+            // Adapted from CpsScmFlowDefinition:
+            Node node = Jenkins.getActiveInstance();
+            FilePath dir;
+            if (job instanceof TopLevelItem) {
+                FilePath baseWorkspace = node.getWorkspaceFor((TopLevelItem) job);
```
amuniz: Is this creating the workspace?

src/main/java/org/jenkinsci/plugins/workflow/multibranch/ReadTrustedStep.java
```diff
@@ -0,0 +1,188 @@
...
+                    delegate.checkout(build, dir, listener, node.createLauncher(listener));
+                    if (!file.exists()) {
+                        throw new AbortException(file + " not found");
+                    }
+                    untrustedFile = file.readToString();
+                }
+                SCMStep delegate = new GenericSCMStep(scmSource.build(head, trusted));
+                delegate.setPoll(true);
+                delegate.setChangelog(true);
+                delegate.checkout(build, dir, listener, node.createLauncher(listener));
```
amuniz: Two checkouts... maybe the `SCMSource` should be used (new APIs needed I think). I would be much more efficient for API accesible SCMs.

amuniz:
The need described in JENKINS-34596 is much more simple than what this is implementing. I guess you are killing two birds with one stone: a) using `readTrusted 'Jenkinsfile'` is a way of checking if you are in a trusted branch, and b) using `readTrusted` before any `load` fixes the flaw if the current trusted branches implementation (which allows to run untrusted code from an untrusted branch).

Right?

src/main/java/org/jenkinsci/plugins/workflow/multibranch/ReadTrustedStep.java
jglick: Locates the workspace. Do not do disk I/O.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/ReadTrustedStep.java
jglick: See the note in JIRA about JENKINS-33273.

jglick:
No, I am killing three birds with this stone! :-) `readTrusted` could indeed be used as a security check prior to `load`. (Alternately you can pass the return value to `evaluate`, as the test demonstrates, though this is less desirable since you do not get Replay support etc.; cf. TODO comment.)

But you could also use it to load any other file which might control sensitive aspects of the build, not necessarily Pipeline Groovy script. For example, if you wanted to keep a Jenkins label selection in a text file in SCM:

```groovy
node(readTrusted 'label.txt') {
  sh 'make'
}
```

where only repository owners are allowed to touch `Jenkinsfile` and `label.txt`; PR submitters could change `Makefile` and other sources, but not patch `label.txt` to say `master` and thus steal secrets from `$JENKINS_HOME` via a patched `Makefile`. Or you might have something like

```groovy
docker.withRegistry(…) {
  docker.image('mycorp/build-tools').inside {
    sh 'make' // inside container, run whatever you want
  }
  readTrusted 'deploy.sh'
  sh './deploy.sh' // outside container, be careful
}
```

kwhetstone:
:bee: We could add some integration tests around this, even if they're manual to start.  I also imagine this would be an interesting way to control locked resources.

amuniz:
Belated :bee:

PR #11, opened 2016-05-24
Refactored to allow alternate implementations besides fixed Jenkinsfile

jglick:
@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/multibranch/AbstractWorkflowMultiBranchProjectFactory.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/multibranch/AbstractWorkflowMultiBranchProjectFactory.java
@@ -1,0 +1,68 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2016 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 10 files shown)


src/main/java/org/jenkinsci/plugins/workflow/multibranch/AbstractWorkflowMultiBranchProjectFactory.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+public abstract class AbstractWorkflowMultiBranchProjectFactory extends MultiBranchProjectFactory.BySCMSourceCriteria {
+
+    @Override protected final WorkflowMultiBranchProject doCreateProject(ItemGroup<?> parent, String name, Map<String,Object> attributes) {
+        WorkflowMultiBranchProject project = new WorkflowMultiBranchProject(parent, name);
+        customize(project);
+        return project;
+    }
+
+    protected void customize(WorkflowMultiBranchProject project) {}
```
amuniz: What is this for?

src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMVar.java
```diff
@@ -75,20 +78,30 @@
...
-            }
-            return scmSource.build(branch.getHead(), revisionAction.getRevision());
+            tip = revisionAction.getRevision();
         } else {
-            return branch.getScm();
+            SCMHead head = branch.getHead();
+            TaskListener listener = TaskListener.NULL; // TODO would like to send to build log, but have no way of doing that here
+            tip = scmSource.fetch(head, listener);
+            if (tip == null) {
+                // TODO would it be better to throw AbortException? Cf. SCMBinder.
```
amuniz: I think so. Otherwise `checkout scm` would not be checking out what you (as user) expect it's checking out.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactory.java
```diff
@@ -25,63 +25,38 @@
...
+    @Extension public static class DescriptorImpl extends AbstractWorkflowBranchProjectFactoryDescriptor {
 
         @Override public String getDisplayName() {
-            return "Fixed configuration";
-        }
-
-        @SuppressWarnings("rawtypes") // erasure
-        @Override public boolean isApplicable(Class<? extends MultiBranchProject> clazz) {
-            return WorkflowMultiBranchProject.class.isAssignableFrom(clazz);
+            return "by " + SCRIPT;
```
amuniz: Where is this shown in the UI? Under project recognizers? If so, shouldn't it start with uppercase.

amuniz:
:bee:

src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMVar.java
jglick: Yes; for purposes of this PR I wanted to avoid making changes that are not really related to the refactoring. You can see the existing behavior in the test for the case of a deleted branch. Probably a failure here (and in the analogous code in `SCMBinder`) would be clearer, assuming there are no _legitimate_ cases where `fetch` would return null here. But best done in a follow-up.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactory.java
jglick: Yes, under project recognizers. Perhaps. Visually it sort of completes the text above it, but it is true that we generally use uppercase style for selectable choices of this kind.

kwhetstone:
:bee:

PR #12, opened 2016-05-25
[JENKINS-30206] [JENKINS-32396] Honor changes to NoTriggerBranchProperty

jglick:
Integration test for https://github.com/jenkinsci/branch-api-plugin/pull/43. Pending https://github.com/jenkinsci/branch-api-plugin/pull/36, fixes a bug whereby changes to the trigger settings on a (single-repository) multibranch project would not take effect on branch projects that had already been created. (Organization folders did not suffer from an analogous bug.)

Supersedes https://github.com/jenkinsci/workflow-multibranch-plugin/pull/2.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactory.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactory.java
@@ -59,7 +59,7 @@ public class WorkflowBranchProjectFactory extends BranchProjectFactory<WorkflowJ
         try {
             if (property == null) {
                 project.addProperty(new BranchJobProperty(branch));
-            } else if (!property.getBranch().equals(branch)) {
+            } else { // TODO may add equality check if https://github.com/jenkinsci/branch-api-plugin/pull/36 or equivalent is implemented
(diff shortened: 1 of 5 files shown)


src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactory.java
```diff
@@ -59,7 +59,7 @@
         try {
             if (property == null) {
                 project.addProperty(new BranchJobProperty(branch));
-            } else if (!property.getBranch().equals(branch)) {
+            } else { // TODO may add equality check if https://github.com/jenkinsci/branch-api-plugin/pull/36 or equivalent is implemented
```
kwhetstone: This might be useful, but I think it's good to push it out to another PR.

kwhetstone:
:bee:

src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactory.java
jglick: It _was_ in a separate PR, which stalled due to some problems described there.

PR #13, opened 2016-06-01
Adding a keepBuilds standalone step.

abayer:
Log rotation is a common enough need that it made sense to have a separate step for it. I'm open to a different name if "keepBuilds" isn't ideal.

I tried to include the config.jelly for LogRotator directly, but the field names threw that off. Other hoops had to be jumped through for the Snippet Generator to work right and for null fields to be handled correctly in instantiating the LogRotator.

cc @reviewbybees @jglick

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStep.java
@@ -91,7 +91,7 @@ public class JobPropertyStep extends AbstractStepImpl {
             BulkChange bc = new BulkChange(job);
             try {
                 for (JobProperty prop : job.getAllProperties()) {
-                    if (prop instanceof BranchJobProperty) {
+                    if (prop instanceof BranchJobProperty || !isConfiguredProperty(prop)) {
                         // TODO do we need to define an API for other properties which should not be removed?
                         continue;
                     }
@@ -107,9 +107,27 @@ public class JobPropertyStep extends AbstractStepImpl {
             return null;
         }
 
+        /**
+         * Checks to see if the existing {@link JobProperty} is of a type defined in the step. We will only remove
+         * properties that are defined in the step.
+         *
+         * @param existingProperty An existing property on the job.
+         * @return True if the existing property is of a type defined in the step, false otherwise
+         */
+        private boolean isConfiguredProperty(JobProperty existingProperty) {
+            for (JobProperty prop : step.properties) {
+                if (prop.getClass().isInstance(existingProperty)) {
+                    return true;
+                }
+            }
+
+            return false;
+        }
+
         private static final long serialVersionUID = 1L;
 
     }
+
 
     @Extension public static class DescriptorImpl extends AbstractStepDescriptorImpl {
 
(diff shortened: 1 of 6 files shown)


michaelneale:
🐝

jglick:
This seems like a slippery slope. People will start expecting other job properties to have their own steps, too.

Also a subsequent `properties` will clobber the setting.

abayer:
Hmm. Interesting point in re properties - that's a problem. But as to the
slippery slope - why not steps for more properties?

michaelneale:
@jglick for important properties this may indeed be required, things that would have once been explicit in freestyle. There is no enumeration of what these properties are or how to set them (and they will never be found in google), hence the need for this. And others. It is a slippery slope to a warm pool.

jglick:
> There is no enumeration of what these properties are or how to set them

_Snippet Generator_ lists all applicable `JobProperty`s in the configuration section for the `properties` step.

abayer:
Build discarding is such a common (near-universal!) configuration that I don't think it makes sense to require knowing to look in `properties` in the Snippet Generator and then copy-pasting that ugly chunk of code to every Jenkinsfile.

abayer:
Updated to only remove `JobProperty`s if they're of a type defined in the `properties` step - so yes, if you call `keepBuilds` and then `properties` with a different configuration of `BuildDiscarderProperty`, the second will override, but if you call `keepBuilds` and then `properties` without `BuildDiscarderProperty`, the one defined by `keepBuilds` will stay.

jglick:
> Updated to only remove `JobProperty`s if they're of a type defined in the `properties` step

Not good—this means if you ever have some `JobProperty` defined in your `Jenkinsfile`, you can never again remove it! (Some `JobProperty`s are of the sort that are expected to always be present, with some sort of flags that effectively disable their behavior; but others, which going forward should be `OptionalJobProperty`s, really need to be absent to get the base behavior—and this category includes `ParametersDefinitionProperty`.)

abayer:
Ok, so...special casing for `BuildDiscarderProperty`, then?

jglick:
And special casing for other `JobProperty`s we decide to add custom steps for in the future? Which would be impossible for steps not present in this plugin, without a new API.

This just seems problematic. If you really want syntactically sweeter log rotation, it would be better to define a step that does not touch `JobProperty`s, and just calls `LogRotator.perform` directly, rather than waiting for it to be called implicitly from `WorkflowRun.finish`. (Unfortunately `BuildDiscarder.perform` does not declare a `TaskListener` argument so we cannot print status to the build log; this would be a simple core API enhancement.)

abayer:
I think it's worth considering reevaluating `properties` if it's this problematic to work around - what's there now may be a sufficient band-aid with `properties` as the only way to interact with `JobProperty`s, but I think that's flawed. That said, I don't have a great idea as an alternative yet.

michaelneale:
oh, I did not realise those properties were enumerable in the properties step (then syntax/usability are the place to fix). 

A step to call log rotator directly may be a reasonable compromise.

abayer:
The con for me of invoking `LogRotator` directly is that we then hit the common issue of making sure users know to execute it in a try/catch/finally or after `catchError`, etc, or running it at the beginning of the build... I also recently ran into a bit of frustration recently when I realized I couldn't just go to the script console and automatically update the log rotation settings for every job on a master once Pipeline jobs were involved, but I guess that's an issue either way. I almost wonder if it might make sense to have some way better way of managing `JobProperty`s (or some subset of them) at a repo or org level...

jglick:
> or running it at the beginning of the build

This is what I would generally recommend.

abayer:
Closing this - I think the work @kohsuke is doing on meta steps is good enough.

PR #14, opened 2016-06-02
Add some useful things from SCMBinderTest to the test JAR

jglick:
Forgotten in #11.

@reviewbybees

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinderTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinderTest.java
@@ -102,7 +102,7 @@ public class SCMBinderTest {
         WorkflowMultiBranchProjectTest.showIndexing(mp);
     }
 
-    static void assertRevisionAction(WorkflowRun build) {
+    public static void assertRevisionAction(WorkflowRun build) {
         SCMRevisionAction revisionAction = build.getAction(SCMRevisionAction.class);
         assertNotNull(revisionAction);
         SCMRevision revision = revisionAction.getRevision();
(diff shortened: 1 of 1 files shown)


abayer:
:bee:

src/test/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinderTest.java
```diff
@@ -102,7 +102,7 @@
         WorkflowMultiBranchProjectTest.showIndexing(mp);
     }
 
-    static void assertRevisionAction(WorkflowRun build) {
+    public static void assertRevisionAction(WorkflowRun build) {
```
oleg-nenashev: Some Javadocs would be useful

src/test/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinderTest.java
```diff
@@ -256,8 +256,8 @@ static void assertRevisionAction(WorkflowRun build) {
         r.assertLogContains("subsequent content", b);
         r.assertLogContains("not trusting", b);
     }
-    static class WarySource extends GitSCMSource {
-        WarySource(String id, String remote, String credentialsId, String includes, String excludes, boolean ignoreOnPushNotifications) {
+    public static class WarySource extends GitSCMSource {
```
oleg-nenashev: doh

svanoort:
🐝

src/test/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinderTest.java
jglick: Well, it is for tests only—if you know you need to use it, you can look at the source and find usages.

PR #17, opened 2016-06-24
Removing Mercurial-specific tests

jglick:
Related to https://github.com/jenkinsci/workflow-scm-step-plugin/pull/4.

@reviewbybees

PR #18, opened 2016-06-24
Demonstrate that changelogs are computed for noninitial builds of Git and Subversion branch projects

jglick:
Note that there is a trivial merge conflict with #17. The corresponding case does _not_ currently pass for Mercurial; under investigation.

@reviewbybees


abayer:
:bee:

PR #20, opened 2016-07-14
JENKINS-34005 - update tests for PipelineTriggersJobProperty

abayer:
[JENKINS-34005](https://issues.jenkins-ci.org/browse/JENKINS-34005)

Depends on https://github.com/jenkinsci/workflow-job-plugin/pull/9

* Updates existing tests to deal with the fact that there should always be a `PipelineTriggersJobProperty` on a `WorkflowJob`, even if it has no triggers in it.
* Added a new test specifically for `PipelineTriggersJobProperty`.

cc @reviewbybees esp @jglick

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,12 +69,12 @@ THE SOFTWARE.
         <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>2.4-SNAPSHOT</version>
+            <version>2.4</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-cps</artifactId>
-            <version>2.10-SNAPSHOT</version>
+            <version>2.10</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
@@ -104,7 +104,7 @@ THE SOFTWARE.
         <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>workflow-basic-steps</artifactId>
-            <version>2.0</version>
+            <version>2.1</version>
             <scope>test</scope>
         </dependency>
         <dependency>
@@ -122,7 +122,7 @@ THE SOFTWARE.
         <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>workflow-step-api</artifactId>
-            <version>2.2</version>
+            <version>2.3</version>
             <classifier>tests</classifier>
             <scope>test</scope>
         </dependency>
@@ -136,7 +136,7 @@ THE SOFTWARE.
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-cps</artifactId>
-            <version>2.10-SNAPSHOT</version>
+            <version>2.10</version>
             <classifier>tests</classifier>
             <scope>test</scope>
         </dependency>
@@ -188,7 +188,7 @@ THE SOFTWARE.
         <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>2.4-SNAPSHOT</version>
+            <version>2.4</version>
             <classifier>tests</classifier>
             <scope>test</scope>
         </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -179,6 +179,13 @@ THE SOFTWARE.
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>workflow-job</artifactId>
+            <version>2.4-SNAPSHOT</version>
+            <classifier>tests</classifier>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.jenkins-ci.modules</groupId>
             <artifactId>sshd</artifactId>
             <version>1.6</version>
--- a/src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
@@ -33,6 +33,9 @@ import hudson.model.queue.QueueTaskFuture;
 import hudson.tasks.LogRotator;
 import java.util.Collections;
 import java.util.List;
+
+import hudson.triggers.TimerTrigger;
+import hudson.triggers.Trigger;
 import jenkins.branch.BranchProperty;
 import jenkins.branch.BranchSource;
 import jenkins.branch.DefaultBranchPropertyStrategy;
@@ -44,6 +47,7 @@ import org.jenkinsci.plugins.scriptsecurity.scripts.ScriptApproval;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
 import org.jenkinsci.plugins.workflow.job.WorkflowJob;
 import org.jenkinsci.plugins.workflow.job.WorkflowRun;
+import org.jenkinsci.plugins.workflow.job.properties.PipelineTriggersJobProperty;
 import org.jenkinsci.plugins.workflow.steps.StepConfigTester;
 import static org.junit.Assert.*;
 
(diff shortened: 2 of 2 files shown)


abayer:
Oh, and I deployed a SNAPSHOT of https://github.com/jenkinsci/workflow-job-plugin/pull/9 so these tests should work.

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
```diff
@@ -185,4 +191,66 @@
...
+
+        // Now add a trigger.
+        p.setDefinition(new CpsFlowDefinition("properties([[$class: 'PipelineTriggersJobProperty',\n"
+                + "  triggers: [[$class: 'TimerTrigger', spec: '@daily']]]])\n"
+                + "echo 'foo'"));
+
+        r.assertBuildStatusSuccess(p.scheduleBuild2(0));
+
+        assertEquals(1, p.getTriggers().size());
+
```
jglick: :bug: you are not verifying that `start` was called with the correct argument.

> calling `startTriggers()` in `PipelineTriggersJobProperty.setOwner` causes...chaos. Way too many calls to `startTriggers()`

Yet this may be the only way you get that call done properly. Why would there be multiple calls to `setOwner`?

abayer: I think I have unraveled that mess. Will dig into the `start` argument.

abayer: Actually, I may not have. Will touch back on this tomorrow!

abayer: Hrm - how do I check the `start` argument? Guess I need to find a `Trigger` that actually cares about the `newInstance` flag - `TimerTrigger` doesn't...

jglick: The upstream PR had such a `Trigger`, no?

abayer:
Retriggering build - last run used a stale snapshot of `workflow-job`.

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
```diff
@@ -185,4 +199,109 @@
...
+        assertTrue(p.getTriggers().isEmpty());
+
+        r.assertBuildStatusSuccess(p.scheduleBuild2(0));
+
+        // Make sure the triggers are still empty.
+        assertTrue(p.getTriggers().isEmpty());
+
+        // Now add a trigger.
+        p.setDefinition(new CpsFlowDefinition("properties([[$class: 'PipelineTriggersJobProperty',\n"
```
jglick: Since there is a `@Symbol`, I guess this will become `pipelineTriggers(…)` soon.

abayer: ...once we've gotten [JENKINS-29922](https://issues.jenkins-ci.org/browse/JENKINS-29922) fully landed, yeah - I think we're still going to want the `properties` step containing `pipelineTriggers(...)` and friends, though, since we wipe out any pre-existing properties currently...

jglick: :ant: This dependency has been released, so please update to it and use it.

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
```diff
@@ -211,4 +225,76 @@
...
+
+        assertNotNull(mockTrigger);
+
+        assertTrue(mockTrigger.isStarted);
+
+        assertEquals("[null, false]", MockTrigger.startsAndStops.toString());
+
+        // Now run a properties step with a different property and verify that we still have a
+        // PipelineTriggersJobProperty, but with no triggers in it.
+        p.setDefinition(new CpsFlowDefinition("properties([[$class: 'DisableConcurrentBuildsJobProperty']])\n"
```
jglick: :ant: ditto, this has a symbol we can use now too.

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
abayer: Will do.

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
abayer: Yup yup.

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
abayer: Note that `TimerTrigger` does *not* have a `@Symbol` on it, so that's gonna have to still be the same.

abayer: And at least for this moment, I'm not going to worry about having a `@Symbol` on `MockTrigger` either.

pom.xml
```diff
@@ -186,6 +186,13 @@ THE SOFTWARE.
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>workflow-job</artifactId>
+            <version>2.4-SNAPSHOT</version>
```
jglick: You can make this 2.4 now.

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
jglick: > TimerTrigger does not have a @Symbol on it

It does in Jenkins 2.2+. So use the switch idiom used in other test cases in this suite, and try running with

    mvn -Djenkins.version=2.7.1 -Dtest=JobPropertyStepTest surefire:test

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
```diff
@@ -211,4 +225,76 @@
...
+        assertTrue(p.getTriggers().isEmpty());
+
+        r.assertBuildStatusSuccess(p.scheduleBuild2(0));
+
+        // Make sure the triggers are still empty.
+        assertTrue(p.getTriggers().isEmpty());
+
+        // Now add a trigger.
+        p.setDefinition(new CpsFlowDefinition("properties([[$class: 'PipelineTriggersJobProperty',\n"
+                + "  triggers: [[$class: 'TimerTrigger', spec: '@daily'],\n"
```
jglick: Should I think wind up looking like

```groovy
properties([
  pipelineTriggers([
    cron('@daily'),
    mock()
  ])
])
```

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
```diff
@@ -81,7 +82,9 @@
         BooleanParameterDefinition.DescriptorImpl.class.isAnnotationPresent(Symbol.class) && // "booleanParam"
         StringParameterDefinition.DescriptorImpl.class.isAnnotationPresent(Symbol.class) && // "string"
         BuildDiscarderProperty.DescriptorImpl.class.isAnnotationPresent(Symbol.class) && // "buildDiscarder"
-        LogRotator.LRDescriptor.class.isAnnotationPresent(Symbol.class); // "logRotator"
+        LogRotator.LRDescriptor.class.isAnnotationPresent(Symbol.class) && // "logRotator"
+        TimerTrigger.DescriptorImpl.class.isAnnotationPresent(Symbol.class) && // "disableConcurrentBuilds"
+        PipelineTriggersJobProperty.DescriptorImpl.class.isAnnotationPresent(Symbol.class); // "pipelineTriggers"
```
jglick: :bug: this last clause is unconditionally true, since we declare that dependency. This constant is about symbols present in _core_.

jglick:
Merge against `master` to pick up #24.

abayer:
Merged.

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
```diff
@@ -211,4 +231,91 @@
         r.assertBuildStatusSuccess(r.waitForCompletion(b4));
     }
 
+    @Issue("JENKINS-34005")
+    @Test public void triggersProperty() throws Exception {
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        // Verify the base case behavior.
+        p.setDefinition(new CpsFlowDefinition("echo 'foo'"));
```
jglick: Actually just `""` suffices.

jglick:
:bee:

PR #21, opened 2016-07-27
More polite error messages

jglick:
@reviewbybees


src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMVar.java
```diff
@@ -63,7 +64,7 @@
         Run<?,?> build = script.$build();
         // TODO some code overlap with SCMBinder.create, but not obvious how to factor out common parts
         if (!(build instanceof WorkflowRun)) {
-            throw new AbortException("not available outside a Pipeline build");
+            throw new AbortException("‘checkout scm’ is not available outside a Pipeline build");
```
abayer: :ant: Not a huge fan of the typeset quotes, but not a dealbreaker.

abayer:
:bee: modulo my distaste for typeset quotes. =)

jglick:
```
java.lang.AssertionError: expected:<[]> but was:<[org.jenkinsci.plugins.workflow.job.properties.PipelineTriggersJobProperty@25d0566a]>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at org.jenkinsci.plugins.workflow.multibranch.JobPropertyStepTest.configRoundTripBuildDiscarder(JobPropertyStepTest.java:85)
```

The PR builder seems to be mixing this up with #20 for reasons I cannot fathom.

abayer:
Seen that sort of thing happen a couple times lately. Is odd. Close/reopen, I guess?

jglick:
Looks like it was a problem with mismatched snapshots.

PR #24, opened 2016-07-28
Bump parent POM to fix PCT against new cores.

alvarolobato:
@reviewbybees


jglick:
:bee:

jglick:
I think the default concurrency setting for the templatized jobs (`1C`) overwhelms the CI builder. Probably the JVM is reporting the number of physical cores whereas the actual agent is running in a VM with many fewer.

jglick:
(but until a recentish update to the parent POM, `concurrency` was not actually honored)

PR #25, opened 2016-08-05
Adding a test of an empty triggers property removing triggers.

abayer:
Just wanted to make sure we had this and working correctly.

cc @reviewbybees esp @jglick

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
@@ -59,6 +59,8 @@ import org.jenkinsci.plugins.workflow.job.properties.PipelineTriggersJobProperty
 import org.jenkinsci.plugins.workflow.properties.MockTrigger;
 import org.jenkinsci.plugins.workflow.steps.StepConfigTester;
 import org.jenkinsci.plugins.workflow.test.steps.SemaphoreStep;
+
+import static org.hamcrest.core.StringContains.containsString;
 import static org.junit.Assert.*;
 
 import org.junit.After;
@@ -268,7 +270,9 @@ public class JobPropertyStepTest {
         r.assertLogContains(Messages.JobPropertyStep__could_remove_warning(), b);
         // Verify that we're not seeing warnings for any properties being removed, since there are no pre-existing ones.
         // Note - not using Messages here because we're not actually removing any properties.
-        r.assertLogNotContains("WARNING: Removing existing job property", b);
+        String warningSubString = "WARNING: Removing existing job property";
+        assertThat(Messages.JobPropertyStep__removed_property_warning(""), containsString(warningSubString));
+        r.assertLogNotContains(warningSubString, b);
 
         assertEquals(2, p.getTriggers().size());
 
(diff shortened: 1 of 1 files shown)


abayer:
Need to add resetting of the start/stop list.

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
```diff
@@ -296,6 +308,64 @@
         assertEquals("[null, false, null]", MockTrigger.startsAndStops.toString());
     }
 
+    @Test public void scmAndEmptyTriggersProperty() throws Exception {
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        // Verify the base case behavior.
+        p.setDefinition(new CpsFlowDefinition("echo 'foo'"));
```
jglick: :ant: always pass an explicit `sandbox` argument, typically `true` for tests unless you specifically need to test the opposite,

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
```diff
@@ -296,6 +308,64 @@
...
+
+        // Make sure the triggers are still empty.
+        assertTrue(p.getTriggers().isEmpty());
+
+        // Now add a trigger.
+        p.setDefinition(new CpsFlowDefinition(
+                (HAVE_SYMBOL ?
+                        "properties([pipelineTriggers([\n"
+                                + "  scm('@daily')])\n" :
+                        // TODO: Make this work with just "spec: '@daily'". Doesn't work for SCMTrigger due to different
```
jglick: Well as of Jenkins 2 we do not care any more anyway.

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
```diff
@@ -296,6 +308,64 @@
...
+                        // names on DataBoundConstructor...
+                        "properties([pipelineTriggers([[$class: 'SCMTrigger', scmpoll_spec: '@daily']])])\n"
+                ) + "echo 'foo'"));
+
+        WorkflowRun b = r.assertBuildStatusSuccess(p.scheduleBuild2(0));
+
+        // Verify that we're seeing warnings due to running 'properties' in a non-multibranch job.
+        r.assertLogContains(Messages.JobPropertyStep__could_remove_warning(), b);
+        // Verify that we're not seeing warnings for any properties being removed, since there are no pre-existing ones.
+        // Note - not using Messages here because we're not actually removing any properties.
```
jglick: :ant: huh? You should still use `Messages` in the test, so that the test will correctly fail if you regress something after editing the message.

jglick:
:bee:

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
abayer: I couldn't match the strings via `Messages` when not actually removing anything - i.e., how do I make sure a `Message` with undefined parameters isn't there? Doesn't really make sense.

jglick: Oh I see what you are saying—the message takes a format parameter, and we do not have a substring for just the generic part of the message.

Still, you could for example write

```java
String substring = "WARNING: Removing existing job property";
assertThat(Messages.JobPropertyStep__removed_property_warning(""), containsString(substring));
r.assertLogNotContains(substring, b);
```

which ensures that your test is not accidentally weakened.

jglick:
re:bee:

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
abayer: I like it! Adding that.

jglick:
re:bee:

PR #26, opened 2016-08-05
[FIXED JENKINS-37219] Allow suppressing builds triggered by indexing

abayer:
[JENKINS-37219](https://issues.jenkins-ci.org/browse/JENKINS-37219)

Specifically, allow this from the Jenkinsfile via a JobProperty.

cc @reviewbybees esp @jglick

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SuppressBranchIndexingBuildsJobProperty.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SuppressBranchIndexingBuildsJobProperty.java
@@ -1,18 +1,25 @@
 /*
- * Licensed to the Apache Software Foundation (ASF) under one or more
- * contributor license agreements.  See the NOTICE file distributed with
- * this work for additional information regarding copyright ownership.
- * The ASF licenses this file to You under the Apache License, Version 2.0
- * (the "License"); you may not use this file except in compliance with
- * the License.  You may obtain a copy of the License at
+ * The MIT License
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ * Copyright 2016 CloudBees, Inc.
  *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
  */
 package org.jenkinsci.plugins.workflow.multibranch;
 
--- a/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SuppressBranchIndexingBuildsJobProperty.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SuppressBranchIndexingBuildsJobProperty.java
@@ -1,0 +1,73 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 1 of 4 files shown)


abayer:
Whoops, forgot config-details...

src/main/java/org/jenkinsci/plugins/workflow/multibranch/SuppressBranchIndexingBuildsJobProperty.java
```diff
@@ -0,0 +1,80 @@
...
+    public SuppressBranchIndexingBuildsJobProperty() {
+    }
+
+    @Extension
+    @Symbol("suppressBranchIndexingBuilds")
+    public static class DescriptorImpl extends OptionalJobPropertyDescriptor {
+
+        @Override public String getDisplayName() {
+            return "Suppress builds triggered by branch indexing";
+        }
```
jglick: Needs to be made applicable only to multibranch projects: makes no sense to display this option for standalone `WorkflowJob`s.

jglick:
Need a `help.html` or whatever describing what it does.

jglick:
So IIUC this allows you to suppress automatic _subsequent_ builds of a branch project which is not suppressed at either the repository or organization level (and so which would have had an automatic _initial_ build, which runs `properties`).

What about the reverse case? I.e., say there is an organization-level blanket suppression of automatic builds, but for a particular repository or just a single branch we _do_ wish to have automatic subsequent builds (and we do not mind manually triggering the initial build). That seems like a reasonable use case, but this design does not allow it. It would seem like a simple change to make the job property keep a `boolean` field for automatic build enablement; when the property is present it would override any repository- or organization-level setting.

Since the other `QueueTaskDispatcher`s are in `branch-api`, that would mean moving the main logic to `branch-api`, and making it an `OptionalJobProperty<Job>`, probably reworking this PR to just be an integration test.

src/test/java/org/jenkinsci/plugins/workflow/multibranch/SuppressBranchIndexingBuildsJobPropertyTest.java
```diff
@@ -0,0 +1,106 @@
...
+        WorkflowRun secondBuild = r.assertBuildStatusSuccess(master.scheduleBuild2(0));
+        assertNotNull(secondBuild);
+        assertEquals(2, secondBuild.getNumber());
+        assertEquals(3, master.getNextBuildNumber());
+        assertNull(master.getProperty(SuppressBranchIndexingBuildsJobProperty.class));
+
+
+        // Now let's see it actually trigger another build from a new commit.
+        round3();
+        WorkflowMultiBranchProjectTest.showIndexing(p);
```
jglick: Really need to make this a stock utility method in `ComputedFolder`.

src/test/java/org/jenkinsci/plugins/workflow/multibranch/SuppressBranchIndexingBuildsJobPropertyTest.java
```diff
@@ -0,0 +1,106 @@
...
+/**
+ * Integration test for {@link SuppressBranchIndexingBuildsJobProperty}
+ */
+@Issue("JENKINS-37219")
+public class SuppressBranchIndexingBuildsJobPropertyTest {
+
+    @Rule public JenkinsRule r = new JenkinsRule();
+    @Rule public GitSampleRepoRule sampleRepo = new GitSampleRepoRule();
+
+    @Issue("JENKINS-37219")
```
jglick: redundant

src/test/java/org/jenkinsci/plugins/workflow/multibranch/SuppressBranchIndexingBuildsJobPropertyTest.java
```diff
@@ -0,0 +1,106 @@
...
+
+        // Should be able to manually build master, which should result in the blocking going away.
+        WorkflowRun secondBuild = r.assertBuildStatusSuccess(master.scheduleBuild2(0));
+        assertNotNull(secondBuild);
+        assertEquals(2, secondBuild.getNumber());
+        assertEquals(3, master.getNextBuildNumber());
+        assertNull(master.getProperty(SuppressBranchIndexingBuildsJobProperty.class));
+
+
+        // Now let's see it actually trigger another build from a new commit.
```
jglick: Test looks good.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/SuppressBranchIndexingBuildsJobProperty.java
abayer: Okiedokie. So just change to `OptionalJobProperty<WorkflowMultiBranchProject>` or is more needed than that?

abayer: Also, we've got `BranchJobProeprty` scoped only to `WorkflowJob`...

abayer:
Good point on the reverse case. Let me think about that.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/SuppressBranchIndexingBuildsJobProperty.java
jglick: > just change to `OptionalJobProperty<WorkflowMultiBranchProject>`

No, `WorkflowMultiBranchProject` is not a `Job`. You need to override (IIRC) `isApplicable`.

abayer: Still kinda confused - how can I tell the difference between a `WorkflowJob` from a `WorkflowMultiBranchProject` or otherwise?

jglick: `getParent() instanceof SCMSource` or `getParent() instanceof MultiBranchProject` is what I would try. I think this idiom is used elsewhere (`SCMVar` IIRC).

jglick: Anyway if you do move this to `branch-api` then you would have `Job` as the property type parameter so `isApplicable` would be the only way this gets hidden from non-multibranch projects.

abayer: Ah, ok. I'll get started on a `branch-api` PR today.

abayer:
Replacing with a test-only pull request.

PR #27, opened 2016-08-17
[JENKINS-37219] Test for new overrideIndexTriggers job property.

abayer:
[JENKINS-37219](https://issues.jenkins-ci.org/browse/JENKINS-37219)

Downstream of https://github.com/jenkinsci/branch-api-plugin/pull/48

cc @reviewbybees esp @jglick

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
@@ -355,7 +355,6 @@ public class JobPropertyStepTest {
         assertNotNull(master.getProperty(OverrideIndexTriggersJobProperty.class));
         assertFalse(master.getProperty(OverrideIndexTriggersJobProperty.class).getEnableTriggers());
 
-        sampleRepo.git("checkout", "master");
         sampleRepo.write("Jenkinsfile", "properties([])");
         sampleRepo.git("commit", "--all", "--message=master-2");
         sampleRepo.notifyCommit(r);
@@ -373,7 +372,6 @@ public class JobPropertyStepTest {
 
 
         // Now let's see it actually trigger another build from a new commit.
-        sampleRepo.git("checkout", "master");
         sampleRepo.write("Jenkinsfile", "// yet more");
         sampleRepo.git("commit", "--all", "--message=master-3");
         sampleRepo.notifyCommit(r);
--- a/src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
@@ -52,6 +52,7 @@ import jenkins.scm.api.SCMHead;
 import jenkins.scm.api.SCMSource;
 import org.jenkinsci.plugins.scriptsecurity.scripts.ScriptApproval;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
+import org.jenkinsci.plugins.workflow.cps.SnippetizerTester;
 import org.jenkinsci.plugins.workflow.job.WorkflowJob;
 import org.jenkinsci.plugins.workflow.job.WorkflowRun;
 import org.jenkinsci.plugins.workflow.job.properties.DisableConcurrentBuildsJobProperty;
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
```diff
@@ -333,6 +335,64 @@ public void noPropertiesWarnings() throws Exception {
...
+        // Should be initial build of master, which sets the job property.
+        WorkflowJob master = WorkflowMultiBranchProjectTest.scheduleAndFindBranchProject(p, "master");
+        r.waitUntilNoActivity();
+        r.assertBuildStatusSuccess(master.getBuildByNumber(1));
+        assertEquals(2, master.getNextBuildNumber());
+
+        assertNotNull(master.getProperty(OverrideIndexTriggersJobProperty.class));
+        assertFalse(master.getProperty(OverrideIndexTriggersJobProperty.class).getEnableTriggers());
+
+        sampleRepo.git("checkout", "master");
```
jglick: Are we not already on the `master` branch?

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
```diff
@@ -333,6 +335,64 @@ public void noPropertiesWarnings() throws Exception {
...
+    @Issue("JENKINS-37219")
+    @Test
+    public void snippetGeneratorOverrideIndexing() throws Exception {
+        String snippetJson = "{'propertiesMap':\n" +
+                "{'stapler-class-bag': 'true', 'jenkins-branch-OverrideIndexTriggersJobProperty': \n" +
+                "{'specified': true, 'enableTriggers': true}},\n" +
+                "'stapler-class': 'org.jenkinsci.plugins.workflow.multibranch.JobPropertyStep',\n" +
+                "'$class': 'org.jenkinsci.plugins.workflow.multibranch.JobPropertyStep'}";
+
+        new SnippetizerTester(r).assertGenerateSnippet(snippetJson, "properties [overrideIndexTriggers(true)]", null);
```
jglick: Until https://github.com/jenkinsci/workflow-cps-plugin/pull/38 is sorted out, I wonder if we should even bother testing the broken behavior.

abayer: Do you mean https://github.com/jenkinsci/workflow-cps-plugin/pull/49? This predated that work...

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
abayer: Copy-pasta noise. Removing.

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
jglick: After all, this is not actually testing the proper operation of `isOwnerMultibranch`. To do that, we would need to enhance `SnippetizerTester` to actually take a `Snippetizer.LocalAction` instance (from `someJobOrFolderOrWhatever.getAction(Snippetizer.LocalAction.class)`), use HtmlUnit to get its index page, select a step specified by `functionName`, run some specified closure to click some UI controls within the context of the nested configuration form, click _Generate Groovy_, and assert that the resulting textarea matches some expected value.

That would also allow `generateSnippetForBuildTrigger` and variants to be more convincing.

abayer: Oh, yeah, this isn't testing `isOwnerMultibranch` at *all* because OW OW OW OW. It's just testing that the generation actually does what it's supposed to. Or not. Y'know.

jglick: > Do you mean jenkinsci/workflow-cps-plugin#49?

Uh, yes, sorry.

jglick:
:bee:

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
jglick: > It's just testing that the generation actually does what it's supposed to.

Well, it is testing that the _serverside_ code is correct assuming the form was correct and _Generate Groovy_ was passing the right JSON back to the server. It does not verify that the right JSON gets sent, or in fact that a reasonable form gets displayed at all.

abayer: True that. Open me a JIRA for figuring out a way to do the enhancements to `SnippetizerTester`, though - that would definitely be handy.

PR #30, opened 2016-08-29
[JENKINS-37731] Tests for proper handling of SCMTrigger in step.

abayer:
[JENKINS-37731](https://issues.jenkins-ci.org/browse/JENKINS-37731)

Note that we're pretty much stuck with SCMTrigger not working right
with the snippet generator pre-core-2.21 (or whenever
https://github.com/jenkinsci/jenkins/pull/2526 lands). SCMTrigger is
just busted vis a vis the Snippet Generator. So what we're doing here
is testing runtime syntax, and snippet generation against core with a
new enough version to have the "pollScm" symbol on SCMTrigger's descriptor.

Also enabled/fixed some snippetizer round trip tests that now can
work, while I was here.

cc @reviewbybees esp @jglick @daniel-beck

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -74,7 +74,7 @@ THE SOFTWARE.
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-cps</artifactId>
-            <version>2.13</version>
+            <version>2.14-SNAPSHOT</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
@@ -100,11 +100,6 @@ THE SOFTWARE.
             <groupId>${project.groupId}</groupId>
             <artifactId>workflow-scm-step</artifactId>
             <version>2.2</version>
-        </dependency>
-        <dependency>
-            <groupId>org.jenkins-ci.plugins</groupId>
-            <artifactId>structs</artifactId>
-            <version>1.5</version>
         </dependency>
         <dependency>
             <groupId>${project.groupId}</groupId>
@@ -141,7 +136,7 @@ THE SOFTWARE.
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-cps</artifactId>
-            <version>2.13</version>
+            <version>2.14-SNAPSHOT</version>
             <classifier>tests</classifier>
             <scope>test</scope>
         </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -74,7 +74,7 @@ THE SOFTWARE.
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-cps</artifactId>
-            <version>2.10</version>
+            <version>2.14-SNAPSHOT</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -74,7 +74,7 @@ THE SOFTWARE.
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-cps</artifactId>
-            <version>2.10</version>
+            <version>2.14-SNAPSHOT</version>
```
jglick: What is this depending on?

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
```diff
@@ -336,12 +337,86 @@ public void resetStartsAndStops() {
         assertTrue(p.getTriggers().isEmpty());
 
         // Now add a trigger.
+        // Looking for core versions 2.21 and later for the proper pollScm symbol, rather than the broken scm symbol.
+        if (SCMTrigger.DescriptorImpl.class.isAnnotationPresent(Symbol.class)
+                && SymbolLookup.getSymbolValue(SCMTrigger.DescriptorImpl.class).iterator().next().equals("pollScm")) {
```
jglick: Would suffice to write

```java
if (SymbolLookup.getSymbolValue(SCMTrigger.class).contains("pollScm")) {
```

right?

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
```diff
@@ -336,12 +337,86 @@ public void resetStartsAndStops() {
...
+
+            p.setDefinition(new CpsFlowDefinition(
+                    "properties([pipelineTriggers([\n"
+                            + "  pollScm(scmpoll_spec: '@daily', ignorePostCommitHooks: true), [$class: 'MockTrigger']])])\n"
```
jglick: I thought we could use just

```groovy
pollScm('@daily')
```

No?

jglick:
:bee:

pom.xml
abayer: Honestly, can't remember. Switching to 2.13 to see if that works...

abayer: Aaaah, new `structs` version. Adding an explicit one and switching this to 2.13.

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
abayer: Indeed. Good catch.

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
abayer: We can! See the test below, where we are - I wanted to cover both the `pollSCM('@daily')` case and where you want to specify `ignorePostCommitHooks`.

abayer:
Ok, now I remember the reason for cps 2.14-SNAPSHOT - [this and friends](https://jenkins.ci.cloudbees.com/job/plugins/job/workflow-multibranch-plugin/org.jenkins-ci.plugins.workflow$workflow-multibranch/179/testReport/junit/org.jenkinsci.plugins.workflow.multibranch/JobPropertyStepTest/snippetGeneratorOverrideIndexing/) - the fixed snippet generator/`properties` stuff. Removing the structs explicit dependency and switching back to cps 2.14-SNAPSHOT.

jglick:
re:bee:

PR #31, opened 2016-09-02
Ability to set the default Jenkinsfile, in case of absence in the version control system

vaimr:
We have 100500 units in the project. Most of them are collected and deploy same. Keep and update scripts in an amount in a version control system is impossible. But it is possible to proceed as follows (in order of priority reduction):
1. Check for Jenkinsfile at checkout.
2. Check for Jenkinsfile in job workspace folder
3. Use a globally defined default Jenkinsfile (plugin config-file-provider https://wiki.jenkins-ci.org/display/JENKINS/Config+File+Provider+Plugin)

At the same time, one does not prevent us to determine the specific configuration in paragraphs 1 and 2. As well as the actual load Jenkinsfile from SCM in paragraph 3

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -72,6 +72,11 @@ THE SOFTWARE.
             <version>2.6</version>
         </dependency>
         <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>config-file-provider</artifactId>
+            <version>2.11</version>
+        </dependency>
+        <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-cps</artifactId>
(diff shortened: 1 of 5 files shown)


jglick:
Seems like this could be a distinct plugin.

vaimr:
It is. I thought to develop the idea further and receive the default Jenkinsfile, say, from the SCM, as it can be done in a conventional pipeline projects.

jglick:
Do not plan to include functionality along these lines in this plugin. Could be maintained separately.

(FYI CloudBees Jenkins Platform includes a feature along these lines.)

vaimr:
Sketch plugin https://github.com/vaimr/workflow-multibranch-def-plugin

PR #34, opened 2016-09-28
Adapting test to run with or without parameter behavior updates

jglick:
After https://github.com/jenkinsci/workflow-cps-plugin/pull/60 + https://github.com/jenkinsci/workflow-job-plugin/pull/24 build parameters come in principally as environment variables, rather than as entries in the Groovy `[Script.]binding`; *defined* parameters can still be accessed as if they were in the binding, due to the fallback to `env`, but `binding` itself does not list them.

That means the original hack for using `ParametersDefinitionProperty` from a multibranch `Jenkinsfile` no longer works. The variant here works with or without those updates; the commented-out syntax is the intended usage going forward (and one of the motivations for making these changes to begin with).

@reviewbybees esp. @abayer


jglick:
The next step is to implement [JENKINS-35698](https://issues.jenkins-ci.org/browse/JENKINS-35698) so that a simple

```groovy
echo "received ${params.myparam}"
```

would work even on the initial build, printing

```
received default value
```

abayer:
Works for me I don't love the `params` variable (mainly out of fear that I might have something using the same variable name somewhere!) but that is what it is, and so, :bee:

PR #35, opened 2016-10-06
Publish class for extend in other plugins

vaimr:


vaimr:
org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java is package local. I want to use it in my plugin.

vaimr:
Is no longer relevant.

vaimr:
I thought inherited from this class in its plug-in, but decided not to do so. https://github.com/jenkinsci/pipeline-multibranch-defaults-plugin/blob/master/src/main/java/org/jenkinsci/plugins/pipeline/multibranch/defaults/DefaultsBinder.java

PR #36, opened 2016-10-10
[JENKINS-35698] Integration test for params fallback behavior

jglick:
Downstream of https://github.com/jenkinsci/workflow-cps-plugin/pull/73. Amends #34.

@reviewbybees esp. @abayer


oleg-nenashev:
:bee:

PR #42, opened 2016-11-09
SCMBinderTest.exactRevisionSubversion failure

jglick:
Seems to fail on CI, though it passes for me locally; switching to SVNKit, which should insulate the test from unpredictable CLI versions.

@reviewbybees


abayer:
:bee:

PR #44, opened 2016-12-15
[JENKINS-34561] Support for configuring the Jenkinsfile location/name

dbut023:
Adds support to configure the location and name of the Jenkinsfile.

I've tested it with both different file names and sticking Jenkinsfile in a subfolder; my test-multibranch repo shows the test scenarios.

One thing I haven't got working is have the descriptor in SCMBinder return the configured script path as part of its display name.


jglick:
IIRC there is already a plugin that does something much like this; would prefer to keep this plugin simple and add extended behaviors there.

kuhnroyal:
I basically have a similar patch in use currently, if there is another plugin that can intgerate here I have not found it. Would be nice though.

jglick:
https://github.com/jenkinsci/pipeline-multibranch-defaults-plugin

leewinder:
What's the expectation of getting this pull request merged in?  The other plug-in (with defaults) doesn't really perform the same behaviour, it simply allows you to specify another (and only one other) file which could be used.

It doesn't allow us to define different names and paths for the Jenkinsfile in branch builds.

Adds a significant and powerful element to this plug-in.

dbut023:
Lee's mirrored my thoughts too. I'm not seeing how this PR slots in with the functionality of multibranch-defaults.

kuhnroyal:
I have to agree, the multibranch-defaults-plugin doesn't solve my problems.

vtintillier:
So is anything that prevents merging this PR?

leewinder:
[workflow-multibranch.hpi.zip](https://github.com/jenkinsci/workflow-multibranch-plugin/files/697331/workflow-multibranch.hpi.zip)

For those needing this, I built @dbut023's repo so I could use the config option.  Just download and unzip the above hpi file, and upload it to Jenkins manually.

Make sure you've installed the official version first as manually uploading it won't sort out the dependancies.  

I added a stupid verison number (100.0.0) so updates to the official release won't override it, but obviously there's no fixes going into that if new official release are done without this pull request being merged in...

y3ti:
Other arguments for merge this patch:

* Normal (not multibranch) pipeline job provides this functionality. We can set "Script Path" and change the default value (Jenkinsfile). 

* There is an issue in Jenkins's Jira: https://issues.jenkins-ci.org/browse/JENKINS-39861

matejsp:
Is there any reason for this to not be merged?
It's the killer feature for multiproduct git repos.

psufoxman:
I will add my vote as well to have this merged.  Would like to have this on the open source.

teramawi:
Voted for this pull request as I am feeling quite limited in multibranch pipeline usage.

Currently I'm using a workaround as mentioned in the [manual loading documentation](https://github.com/jenkinsci/pipeline-plugin/blob/master/TUTORIAL.md#triggering-manual-loading) but with the pipeline model definition coming up this workaround does no longer work.

One cannot use the groovy dsl to load another Jenkinsfile and use the model definition dsl there.
I'd greatly appreciate it, if this feature could find its way into the plugin.

malinoff:
I can describe an other use-case: I'd like to keep my project root simple and clean, all dirty ops-related stuff goes into `ops-tools` directory (including `Jenkinsfile`, of course).

dedalusj:
Any chance this pull request is going to be merged? 

There have been a few use cases already outlined in the comments.

konetzed:
Is there anyway with this add on to have jenkins create multiple jobs based off a directory structure like below

repo/dir1
        dir2
        dir3 
        dirN

Where each dir would have its own Jenkinsfile but the branch would still be just master? 

If not does someone know a better way of doing that?
Thanks

ajohnstone:
@konetzed that would be awesome, this is one of the larger issues I have with using jenkinsfiles. A single repo needing multiple jobs. I think you can generate new jobs from a jenkinsfile.

zloylos:
This is very important and useful feature for many people.
@jglick can u merge it, please?

src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactory.java
```diff
@@ -36,48 +35,12 @@
...
-                return getClass().hashCode();
-            }
-
-            @Override
-            public boolean equals(Object obj) {
-                return getClass().isInstance(obj);
-            }
-        };
+    @DataBoundConstructor public WorkflowBranchProjectFactory() {
+        super(SCRIPT);
```
jglick: Does not make any sense to have two implementation classes.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProject.java
```diff
@@ -56,20 +57,27 @@
...
     }
 
     @Override public SCMSourceCriteria getSCMSourceCriteria(SCMSource source) {
-        return ((AbstractWorkflowBranchProjectFactory) getProjectFactory()).getSCMSourceCriteria(source);
+        return ((AbstractWorkflowBranchProjectFactory)this.getProjectFactory()).getSCMSourceCriteria(source);
     }
 
     @Extension public static class DescriptorImpl extends MultiBranchProjectDescriptor implements IconSpec {
 
+        @DataBoundConstructor public DescriptorImpl(){
```
jglick: Does not make any sense.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProjectFactory.java
```diff
@@ -36,20 +36,36 @@
...
+        return factory.getSCMSourceCriteria(source);
     }
 
     @Extension public static class DescriptorImpl extends MultiBranchProjectFactoryDescriptor {
 
+        private AbstractWorkflowBranchProjectFactory factory;
+
+        public DescriptorImpl(){}
+
+        @DataBoundConstructor public DescriptorImpl(AbstractWorkflowBranchProjectFactory factory){
```
jglick: Ditto.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProject.java
```diff
@@ -56,20 +57,27 @@
 
     private static final Logger LOGGER = Logger.getLogger(WorkflowMultiBranchProject.class.getName());
 
-    public WorkflowMultiBranchProject(ItemGroup parent, String name) {
+    @DataBoundConstructor
```
jglick: This is also wrong.

marvinthepa: Hello @jglick: would it be possible to elaborate on how it should be handled instead?

dbut023:
Closing this PR in preference to PR#59

PR #45, opened 2016-12-16
[JENKINS-39355 follow up] Adapting to the SCM and Branch 2.0.x APIs

stephenc:


stephenc:
@reviewbybees

rsandell:
quick :bee:

src/main/java/org/jenkinsci/plugins/workflow/multibranch/BranchJobProperty.java
```diff
@@ -66,11 +67,17 @@ synchronized void setBranch(@Nonnull Branch branch) {
             @Override public boolean hasPermission(Authentication a, Permission permission) {
                 // This project is managed by its parent and may not be directly configured or deleted.
                 // Note that Item.EXTENDED_READ may still be granted, so you can still see Snippet Generator, etc.
-                if (ACL.SYSTEM.equals(a)) {
+                if (branch instanceof Branch.Dead && (permission == Job.BUILD)) {
+                    // if dead, nobody, not even SYSTEM may build this job
```
jglick: Is this not better covered by `isBuildable`?

src/main/java/org/jenkinsci/plugins/workflow/multibranch/BranchJobProperty.java
```diff
@@ -66,11 +67,17 @@ synchronized void setBranch(@Nonnull Branch branch) {
...
+                    return false;
+                } else if (ACL.SYSTEM.equals(a)) {
                     return true; // e.g., DefaultDeadBranchStrategy.runDeadBranchCleanup
```
jglick: Is this not now covered by the check below? I guess it is safest to unconditionally allow `SYSTEM` to do what it needs, but would be advisable to update comments to reflect the new reality.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/BranchJobProperty.java
```diff
@@ -66,11 +67,17 @@ synchronized void setBranch(@Nonnull Branch branch) {
...
-                } else if (permission == Item.CONFIGURE || permission == Item.DELETE) {
+                } else if (permission == Item.CONFIGURE) {
                     return false;
-                } else {
+                } else if (permission == Item.DELETE) {
+                    // allow early manual clean-up of dead branches
+                    return branch instanceof Branch.Dead && acl.hasPermission(a, permission);
```
jglick: Perhaps more intuitively:

```java
} else if (permission == Item.DELETE && !(branch instanceof Branch.Dead)) {
    return false;
```

else we fall through to the normal ACL check.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactory.java
```diff
@@ -48,7 +49,33 @@
...
+                        if (stat.getAlternativePath() != null) {
+                            listener.getLogger().format("      ‘%s’ not found (but found ‘%s’, search is case sensitive)%n", SCRIPT, stat.getAlternativePath());
+                        } else {
+                            listener.getLogger().format("      ‘%s’ not found%n",SCRIPT);
+                        }
+                        return false;
+                    case DIRECTORY:
+                        listener.getLogger().format("      ‘%s’ found but is a directory not a file%n", SCRIPT);
+                        return false;
+                    default:
```
jglick: Should this not be checking `FILE` only?

src/main/java/org/jenkinsci/plugins/workflow/multibranch/BranchJobProperty.java
stephenc: and how exactly do you get there from here?

src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactory.java
stephenc: It is checking `REGULAR_FILE`, `LINK` and `OTHER` as these should be file-like enough

rsandell:
:bee:

src/main/java/org/jenkinsci/plugins/workflow/multibranch/BranchJobProperty.java
jglick: cf. #53

PR #46, opened 2017-01-09
[FIXED JENKINS-40906] Resolve an SCM from an SCMSource and a list of candidate target branch names

stephenc:
See [JENKINS-40906](https://issues.jenkins-ci.org/browse/JENKINS-40906) 

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/multibranch/ResolveScmStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/multibranch/ResolveScmStep.java
@@ -1,0 +1,290 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/workflow/multibranch/ResolveScmStep.java
```diff
@@ -0,0 +1,297 @@
...
+import org.jenkinsci.plugins.workflow.steps.Step;
+import org.jenkinsci.plugins.workflow.steps.StepContext;
+import org.jenkinsci.plugins.workflow.steps.StepExecution;
+import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.DataBoundSetter;
+
+/**
+ * Resolves an {@link SCM} from a {@link SCMSource} using a priority list of target branch names.
+ *
+ * @since FIXME
```
jglick: Unnecessary in plugins like this; not an API.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/ResolveScmStep.java
```diff
@@ -0,0 +1,297 @@
...
+    public void setIgnoreErrors(boolean ignoreErrors) {
+        this.ignoreErrors = ignoreErrors;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public StepExecution start(StepContext context) throws Exception {
+        return new Execution(context, this);
```
jglick: Do not mix Guiceless `Step` with Guiceful `AbstractStep{Execution,Descriptor}Impl`.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/ResolveScmStep.java
```diff
@@ -0,0 +1,297 @@
...
+                "source=" + source +
+                ", targets=" + targets +
+                ", ignoreErrors=" + ignoreErrors +
+                '}';
+    }
+
+    /**
+     * Our {@link Descriptor}.
+     */
+    @Extension(optional = true)
```
jglick: Do not use optional extensions.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/ResolveScmStep.java
```diff
@@ -0,0 +1,297 @@
...
+        }
+
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        protected SCM run() throws Exception {
+            StepContext context = getContext();
+            TaskListener listener = context.get(TaskListener.class);
+            assert listener != null;
```
jglick: If using Guice (deprecated), use `@StepContextParameter`. Otherwise you must implement `getRequiredContext`.

jglick:
Tip: need to upgrade `workflow-step-api` dep and solve deprecation warnings.

stephenc:
8 days no comments despite asking repeatedly

PR #47, opened 2017-01-13
[JENKINS-33273] Use SCMFileSystem where available

jglick:
[JENKINS-33273](https://issues.jenkins-ci.org/browse/JENKINS-33273)

- [X] optimize `Jenkinsfile` checkout from a branch project
- [X] optimize `CpsScmFlowDefinition` checkout from a standalone project: https://github.com/jenkinsci/workflow-cps-plugin/pull/97
- [x] optimize `readTrusted` checkout from a branch project
- [x] optimize `readTrusted` checkout from a standalone project
- [ ] sanity check using GitHub

Tests take advantage of https://github.com/jenkinsci/git-plugin/pull/455.

@reviewbybees esp. @stephenc

PR #49, opened 2017-01-17
[JENKINS-33273] Use SCMFileSystem where available

jglick:
#47 refiled.

@reviewbybees


jglick:
@reviewbybees done

(unless @stephenc wants to take a look)

PR #53, opened 2017-02-06
[JENKINS-40521] Mark orphaned branch projects as unbuildable

jglick:
[JENKINS-40521](https://issues.jenkins-ci.org/browse/JENKINS-40521)

Downstream of https://github.com/jenkinsci/workflow-job-plugin/pull/37.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/multibranch/BranchJobProperty.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/multibranch/BranchJobProperty.java
@@ -66,7 +66,9 @@ public class BranchJobProperty extends WorkflowJobProperty {
             @Override public boolean hasPermission(Authentication a, Permission permission) {
                 // This project is managed by its parent and may not be directly configured or deleted by users.
                 // Note that Item.EXTENDED_READ may still be granted, so you can still see Snippet Generator, etc.
-                if (permission == Item.CONFIGURE) {
+                if (ACL.SYSTEM.equals(a)) {
+                    return true; // e.g., ComputedFolder.updateChildren
+                } else if (permission == Item.CONFIGURE) {
                     return false;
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/workflow/multibranch/BranchJobProperty.java
```diff
@@ -67,23 +66,25 @@ synchronized void setBranch(@Nonnull Branch branch) {
             @Override public boolean hasPermission(Authentication a, Permission permission) {
                 // This project is managed by its parent and may not be directly configured or deleted by users.
                 // Note that Item.EXTENDED_READ may still be granted, so you can still see Snippet Generator, etc.
-                if (branch instanceof Branch.Dead && (permission == Job.BUILD)) {
-                    // if dead, nobody, not even SYSTEM may build this job
-                    return false;
```
jglick: Only effective against code paths which do a permission check; for example, `scheduleBuild` overloads (called, e.g., from `TimerTrigger`) do not. `isBuildable` is checked more broadly.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/BranchJobProperty.java
```diff
@@ -67,23 +66,25 @@ synchronized void setBranch(@Nonnull Branch branch) {
...
-                } else if (ACL.SYSTEM.equals(a)) {
-                    return true; // e.g., DefaultDeadBranchStrategy.runDeadBranchCleanup
```
jglick: As noted in my comment in #45, this clause (now actually referring to `ComputedFolder.updateChildren`) is obsolete due to the exclusion of `Branch.Dead` below. The `deletedBranch` test confirms it.

jglick: …or not.

PR #54, opened 2017-02-13
Allow implementers to throw IOException

batmat:
@reviewbybees esp. @jglick 

downstream of https://github.com/jenkinsci/branch-api-plugin/pull/82

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -102,7 +102,8 @@ THE SOFTWARE.
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>branch-api</artifactId>
-            <version>2.0.2</version>
+            <!-- TODO: Remove when branch-api is released -->
+            <version>2.0.5-20170213.115501-2</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -102,8 +102,7 @@ THE SOFTWARE.
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>branch-api</artifactId>
-            <!-- TODO: Remove when branch-api is released -->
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -102,7 +102,8 @@ THE SOFTWARE.
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>branch-api</artifactId>
-            <version>2.0.2</version>
+            <!-- TODO: Remove when branch-api is released -->
+            <version>2.0.5-20170213.115501-2</version>
```
jglick: Is this downstream of some other PR? If so, mention it in this PR’s description, and in this comment.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/AbstractWorkflowMultiBranchProjectFactory.java
```diff
@@ -56,7 +56,7 @@
         } // otherwise got recognized by something else before, oh well
     }
 
-    protected void customize(WorkflowMultiBranchProject project) {}
+    protected void customize(WorkflowMultiBranchProject project) throws IOException {}
```
jglick: May as well let it throw `InterruptedException` too.

pom.xml
batmat: downstream of https://github.com/jenkinsci/branch-api-plugin/pull/82

src/main/java/org/jenkinsci/plugins/workflow/multibranch/AbstractWorkflowMultiBranchProjectFactory.java
batmat: @jglick fixed.

batmat:
Just updated the PR to now reference the just released 2.0.7 @jglick

PR #57, opened 2017-02-28
SCMFileSystem kill switch

jglick:
Amends #49 to allow people to disable the new functionality without downgrading, if there is some kind of problem. Cf. https://github.com/jenkinsci/workflow-cps-plugin/pull/97 where there is a UI option.

@reviewbybees


stephenc:
:bee:

PR #59, opened 2017-03-10
[JENKINS-34561] Make Jenkinsfile filename configurable

marvinthepa:
[JENKINS-34561](https://issues.jenkins-ci.org/browse/JENKINS-34561)

This is a re-worked version of #44, hopefully making @jglick happier (although I must confess I do not *completely* understand the issues he had with #44).

Of course I am happy to adjust this further. We could really use this feature, so I will definitely try to respond quickly to suggestions.

@dbut023 this is heavily based on your work, thanks for that.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java
@@ -41,6 +41,7 @@ import jenkins.scm.api.SCMHead;
 import jenkins.scm.api.SCMRevision;
 import jenkins.scm.api.SCMRevisionAction;
 import jenkins.scm.api.SCMSource;
+import org.apache.commons.lang.StringUtils;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
 import org.jenkinsci.plugins.workflow.cps.CpsScmFlowDefinition;
 import org.jenkinsci.plugins.workflow.flow.FlowDefinition;
@@ -59,8 +60,17 @@ class SCMBinder extends FlowDefinition {
     static /* not final */ boolean USE_HEAVYWEIGHT_CHECKOUT = Boolean.getBoolean(SCMBinder.class.getName() + ".USE_HEAVYWEIGHT_CHECKOUT"); // TODO 2.4+ use SystemProperties
     private String scriptPath;
 
+    public Object readResolve() {
+        this.scriptPath = WorkflowBranchProjectFactory.SCRIPT;
+        return this;
+    }
+
     public SCMBinder(String scriptPath) {
-        this.scriptPath = scriptPath;
+        if (StringUtils.isEmpty(scriptPath)) {
+            this.scriptPath = WorkflowBranchProjectFactory.SCRIPT;
+        } else {
+            this.scriptPath = scriptPath;
+        }
     }
 
     @Override public FlowExecution create(FlowExecutionOwner handle, TaskListener listener, List<? extends Action> actions) throws Exception {
--- a/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java
@@ -49,7 +49,6 @@ import org.jenkinsci.plugins.workflow.flow.FlowExecution;
 import org.jenkinsci.plugins.workflow.flow.FlowExecutionOwner;
 import org.jenkinsci.plugins.workflow.job.WorkflowJob;
 import org.jenkinsci.plugins.workflow.job.WorkflowRun;
-import org.kohsuke.stapler.DataBoundConstructor;
 
 /**
  * Checks out the desired version of the script referred to by scriptPath.
--- a/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java
@@ -52,7 +52,7 @@ import org.jenkinsci.plugins.workflow.job.WorkflowRun;
 import org.kohsuke.stapler.DataBoundConstructor;
 
 /**
- * Checks out the desired version of {@link WorkflowBranchProjectFactory#SCRIPT}.
+ * Checks out the desired version of the script referred to by scriptPath.
  */
 class SCMBinder extends FlowDefinition {
 
@@ -114,19 +114,9 @@ class SCMBinder extends FlowDefinition {
     }
 
     @Extension public static class DescriptorImpl extends FlowDefinitionDescriptor {
-        private String scriptPath;
-
-        public DescriptorImpl(){}
-
-        @DataBoundConstructor
-        public DescriptorImpl(String scriptPath){ this.scriptPath = scriptPath; }
-
-        public String getScriptPath(){
-            return scriptPath;
-        }
 
         @Override public String getDisplayName() {
-            return "Pipeline from  + multibranch configuration"; // TODO Messages
+            return "Pipeline from multibranch configuration";
         }
 
     }
--- a/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java
@@ -58,19 +58,17 @@ class SCMBinder extends FlowDefinition {
 
     /** Kill switch for JENKINS-33273 in case of problems. */
     static /* not final */ boolean USE_HEAVYWEIGHT_CHECKOUT = Boolean.getBoolean(SCMBinder.class.getName() + ".USE_HEAVYWEIGHT_CHECKOUT"); // TODO 2.4+ use SystemProperties
-    private String scriptPath;
+    private String scriptPath = WorkflowBranchProjectFactory.SCRIPT;
 
     public Object readResolve() {
-        this.scriptPath = WorkflowBranchProjectFactory.SCRIPT;
+        if (this.scriptPath == null) {
+            this.scriptPath = WorkflowBranchProjectFactory.SCRIPT;
+        }
         return this;
     }
 
     public SCMBinder(String scriptPath) {
-        if (StringUtils.isEmpty(scriptPath)) {
-            this.scriptPath = WorkflowBranchProjectFactory.SCRIPT;
-        } else {
-            this.scriptPath = scriptPath;
-        }
+        this.scriptPath = scriptPath;
     }
 
     @Override public FlowExecution create(FlowExecutionOwner handle, TaskListener listener, List<? extends Action> actions) throws Exception {
--- a/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java
@@ -49,6 +49,7 @@ import org.jenkinsci.plugins.workflow.flow.FlowExecution;
 import org.jenkinsci.plugins.workflow.flow.FlowExecutionOwner;
 import org.jenkinsci.plugins.workflow.job.WorkflowJob;
 import org.jenkinsci.plugins.workflow.job.WorkflowRun;
+import org.kohsuke.stapler.DataBoundConstructor;
 
 /**
  * Checks out the desired version of {@link WorkflowBranchProjectFactory#SCRIPT}.
@@ -57,6 +58,11 @@ class SCMBinder extends FlowDefinition {
 
     /** Kill switch for JENKINS-33273 in case of problems. */
     static /* not final */ boolean USE_HEAVYWEIGHT_CHECKOUT = Boolean.getBoolean(SCMBinder.class.getName() + ".USE_HEAVYWEIGHT_CHECKOUT"); // TODO 2.4+ use SystemProperties
+    private String scriptPath;
+
+    public SCMBinder(String scriptPath) {
+        this.scriptPath = scriptPath;
+    }
 
     @Override public FlowExecution create(FlowExecutionOwner handle, TaskListener listener, List<? extends Action> actions) throws Exception {
         Queue.Executable exec = handle.getExecutable();
@@ -88,8 +94,8 @@ class SCMBinder extends FlowDefinition {
                 if (fs != null) { // JENKINS-33273
                     String script = null;
                     try {
-                        script = fs.child(WorkflowBranchProjectFactory.SCRIPT).contentAsString();
-                        listener.getLogger().println("Obtained " + WorkflowBranchProjectFactory.SCRIPT + " from " + rev);
+                        script = fs.child(scriptPath).contentAsString();
+                        listener.getLogger().println("Obtained " + scriptPath + " from " + rev);
                     } catch (IOException | InterruptedException x) {
                         listener.error("Could not do lightweight checkout, falling back to heavyweight").println(Functions.printThrowable(x).trim());
                     }
@@ -104,13 +110,23 @@ class SCMBinder extends FlowDefinition {
             // Build might fail later anyway, but reason should become clear: for example, branch was deleted before indexing could run.
             scm = branch.getScm();
         }
-        return new CpsScmFlowDefinition(scm, WorkflowBranchProjectFactory.SCRIPT).create(handle, listener, actions);
+        return new CpsScmFlowDefinition(scm, scriptPath).create(handle, listener, actions);
     }
 
     @Extension public static class DescriptorImpl extends FlowDefinitionDescriptor {
+        private String scriptPath;
+
+        public DescriptorImpl(){}
+
+        @DataBoundConstructor
+        public DescriptorImpl(String scriptPath){ this.scriptPath = scriptPath; }
+
+        public String getScriptPath(){
+            return scriptPath;
+        }
 
         @Override public String getDisplayName() {
-            return "Pipeline from " + WorkflowBranchProjectFactory.SCRIPT;
+            return "Pipeline from  + multibranch configuration"; // TODO Messages
         }
 
     }
--- a/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactory.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactory.java
@@ -33,6 +33,7 @@ import jenkins.scm.api.SCMSourceCriteria;
 import org.apache.commons.lang.StringUtils;
 import org.jenkinsci.plugins.workflow.flow.FlowDefinition;
 import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.DataBoundSetter;
 
 /**
  * Recognizes and builds {@code Jenkinsfile}.
@@ -41,11 +42,12 @@ public class WorkflowBranchProjectFactory extends AbstractWorkflowBranchProjectF
     static final String SCRIPT = "Jenkinsfile";
     private String scriptPath = SCRIPT;
 
-    public WorkflowBranchProjectFactory() {
+    @DataBoundConstructor public WorkflowBranchProjectFactory() {
         this.scriptPath = SCRIPT;
     }
 
-    @DataBoundConstructor public WorkflowBranchProjectFactory(String scriptPath) {
+    @DataBoundSetter
+    public void setScriptPath(String scriptPath) {
         if (StringUtils.isEmpty(scriptPath)) {
             this.scriptPath = SCRIPT;
         } else {
@@ -56,7 +58,6 @@ public class WorkflowBranchProjectFactory extends AbstractWorkflowBranchProjectF
     public String getScriptPath(){
         return scriptPath;
     }
-
 
     @Override protected FlowDefinition createDefinition() {
         return new SCMBinder(scriptPath);
@@ -97,17 +98,6 @@ public class WorkflowBranchProjectFactory extends AbstractWorkflowBranchProjectF
     }
 
     @Extension public static class DescriptorImpl extends AbstractWorkflowBranchProjectFactoryDescriptor {
-
-        private String scriptPath = SCRIPT;
-
-        public DescriptorImpl(){ this.scriptPath = SCRIPT; }
-
-        @DataBoundConstructor
-        public DescriptorImpl(String scriptPath){ this.scriptPath = scriptPath; }
-
-        public String getScriptPath(){
-            return scriptPath;
-        }
 
         @Override public String getDisplayName() {
             return "by " + SCRIPT;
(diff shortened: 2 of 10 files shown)


dbut023:
Seems we both pushed more or less the same end result near on simultaneously, although I've somehow ended up with a issue on one file; so this PR is probably got a cleaner history to be integrated into the main repo.

 I did include a unit test (See WorkflowMultiBranchProjectTest)  which is probably worth bringing over into PR.

marvinthepa:
Yes, funny. I will pick your test and documentation into this pull request
by Monday..

myoung34:
Is there anything we can do to help push this?

marvinthepa:
@myoung34: I guess all we can do now is to wait for @jglick to review it again - and hope that he likes it this time.

src/main/resources/org/jenkinsci/plugins/workflow/multibranch/ReadTrustedStep/help.html
```diff
@@ -3,6 +3,6 @@
     Unlike the <code>readFile</code> step, no workspace is required.
     If the associated branch is not trusted, yet the file has been modified from its trusted version, an error is thrown.
     Thus this step is useful for loading scripts or other files which might otherwise be used to run malicious commands.
-    Like <code>checkout scm</code>, as a convenience it may also be used from a standalone project configured with <i>Pipeline from SCM</i>,
```
dbut023: New standard is pipeline with script dropped

marvinthepa:
@dbut023 : right - I thought you had done that on purpose, so I just copied without checking.

I force-pushed a new commit, correctly setting you as the author (as you are the author of these lines).

spiralfuzet:
How cool! We need this in our team. I also like to push this.

PandarinDev:
We also need this fix for our workflow. It would be great if you could review and possibly merge this PR in the near future. Thank you.

matejsp:
@jglick can you merge it, please?
I don't want this MR to be left for months like #44.
Or at least comment if there is anything wrong with the implementation.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactory.java
```diff
@@ -30,39 +30,55 @@
...
 public class WorkflowBranchProjectFactory extends AbstractWorkflowBranchProjectFactory {
-    
     static final String SCRIPT = "Jenkinsfile";
+    private String scriptPath = SCRIPT;
+
+    public WorkflowBranchProjectFactory() {
+        this.scriptPath = SCRIPT;
+    }
+
+    @DataBoundConstructor public WorkflowBranchProjectFactory(String scriptPath) {
```
jglick: no, use `@DataBoundSetter`

src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactory.java
```diff
@@ -82,6 +98,17 @@ public boolean equals(Object obj) {
 
     @Extension public static class DescriptorImpl extends AbstractWorkflowBranchProjectFactoryDescriptor {
 
+        private String scriptPath = SCRIPT;
+
+        public DescriptorImpl(){ this.scriptPath = SCRIPT; }
+
+        @DataBoundConstructor
```
jglick: No, a `Descriptor` is not a `Describable` and so it must not have a `@DataBoundConstructor`.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactory.java
marvinthepa: Thanks for the hint, will do.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactory.java
marvinthepa: Again, thanks for the hint. I am assuming that the descriptor does not actually need to have any references to `scriptPath` - don't know if that is correct. I could not make any sense of the resources that I was able to find about `hudson.model.Descriptor` - at least regarding this case.

matejsp:
@jglick are you now satisfied with the quality? Can we move forward?

marvinthepa:
@matejsp I think we should not pressure jglick, he is probably very busy.

It must be very annoying to review pull requests from total Jenkins plugin newbies like me.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactory.java
```diff
@@ -30,39 +30,56 @@
...
  * Recognizes and builds {@code Jenkinsfile}.
  */
 public class WorkflowBranchProjectFactory extends AbstractWorkflowBranchProjectFactory {
-    
     static final String SCRIPT = "Jenkinsfile";
+    private String scriptPath = SCRIPT;
 
-    @DataBoundConstructor public WorkflowBranchProjectFactory() {}
+    @DataBoundConstructor public WorkflowBranchProjectFactory() {
+        this.scriptPath = SCRIPT;
```
jglick: Redundant initializer.

marvinthepa: fixed.

marvinthepa:
@jglick:

I didn't know about the organization folder. Of course it makes sense to make it configurable on that level as well. I am absolutely willing to add this.

But I would much prefer if (as you suggested), this pull request could make it into the plugin first, with organization support being added in a later release.

ajohnstone:
@marvinthepa with the GitHub organisation, I think it would be best to have a set of paths that can be matched.

And also a file that can be placed in each repo to state, which jenkinsfiles to match. Some repos may need to deviate from a standard convention.

dbut023:
For the organisation folders I do think this will need some additional thought as I'm not sure (something I'm about to look into) if they're setup to handle a repo being in the org folder more than once.

jglick:
I gave it a bit more thought and actually I think it should be very simple: just add a script property (using `@DataBoundSetter`, defaulting again to `Jenkinsfile`) to `WorkflowMultiBranchProjectFactory`; it would then pass on that script name to the repository level. This works because you can have multiple project recognizers in an organization folder. The first to match on a given repository will define the script to use for that `WorkflowMultiBranchProject`. So if you do have a need for different filenames across the organization, you just add a recognizer for each case, and it should all be sorted out automatically.

I will need to do some experimentation with all this in `github-branch-source-plugin/demo/` to make sure the UI works sensibly etc.

marvinthepa:
@jglick is this what you had in mind?

marvinthepa:
Ah, thanks for the hints - I begin to understand a bit better how this all works together.

I added the config.jelly and started manually testing it with github branch source. Setting the script path works, and the scan looks promising:

    Checking branch kevin
      ‘Jenkinsfile.groovy’ found
    Met criteria

However, it is not the branch "kevin" that is built, but "master" instead - which contains `Jenkinsfile`, not `Jenkinsfile.groovy` (see https://github.com/marvinthepa/horst).

I stepped through it a bit, and this seems to be the culprit:

[WorkflowMultiBranchProject.java#L64](https://github.com/jenkinsci/workflow-multibranch-plugin/blob/c7f9ff539397fe5fbe61db51847627db46bf0b4a/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProject.java#L64)

I.e. the constructor is called directly, so the scriptPath is not passed.
I will try to find out what is the right way to fix this. Any hints appreciated.

marvinthepa:
I came up with a workaround, overwrite the `workflowProjectFactory` in the `customize` method - see fbc3fa1.

This "works" when setting a single `WorkflowMultiBranchProjectFactory` per project.

However, using two different project recognizers (as you suggested) does not work for me - 
only the first specified WorkflowMultiBranchProjectFactory is ever instantiated.
I wasn't able to find out if this is because of my wrong implementation, or because of some other (potential) bug.

I will try to find out more, but as always, any hints are highly appreciated.

jglick:
> My test case actually was using two different filenames in different branches of the same repository - I don't know if that is supposed to be supported

No.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProjectFactory.java
```diff
@@ -30,16 +30,28 @@
...
 import jenkins.scm.api.SCMSourceCriteria;
 import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.DataBoundSetter;
+import java.io.IOException;
 
 /**
  * Defines organization folders by {@link WorkflowBranchProjectFactory}.
  */
 public class WorkflowMultiBranchProjectFactory extends AbstractWorkflowMultiBranchProjectFactory {
+    private WorkflowBranchProjectFactory workflowBranchProjectFactory;
```
jglick: No, you do not want to do this. Revert so that the field is `scriptPath`.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProjectFactory.java
```diff
@@ -30,16 +30,28 @@
...
+
+    public String getScriptPath() { return workflowBranchProjectFactory.getScriptPath(); }
+
+    @DataBoundConstructor public WorkflowMultiBranchProjectFactory() {
+        workflowBranchProjectFactory = new WorkflowBranchProjectFactory();
+    }
 
     @Override protected SCMSourceCriteria getSCMSourceCriteria(SCMSource source) {
-        return new WorkflowBranchProjectFactory().getSCMSourceCriteria(source);
+        return workflowBranchProjectFactory.getSCMSourceCriteria(source);
```
jglick: And here you would create a new `WorkflowBranchProjectFactory` with that `scriptPath`, as you had before…

src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProjectFactory.java
```diff
@@ -54,4 +66,8 @@
 
     }
 
+    @Override
+    protected void customize(WorkflowMultiBranchProject project) throws IOException, InterruptedException {
+        project.setProjectFactory(workflowBranchProjectFactory);
```
jglick: But here you also create a new factory with the specified `scriptPath`.

marvinthepa:
Right, config looks much better this way. It was just stupid me trying to get rid of some duplicated code.

spiralfuzet:
Could you merge this feature? I'm really looking forward to use it.
Guys, thanks for the awesome work so far.

sophia-guo:
Agreed with @spiralfuzet .  This is really helpful!

smipi1:
@jglick Are you still planning to merge this? If so, when do you think it will be released for use?

yurigo79:
@jglick - it would be great to merge it. This is a very useful feature for our development flow

verbruggenalex:
Does anyone know of documentation to install this plugin patched? I have a demo next week with this functionality as a key feature. I dont want to pressure maintainers. But it would be a lifesaver if anyone could point me to some docs on how to patch a plugin...

myoung34:
@verbruggenalex `mvn install` https://wiki.jenkins-ci.org/display/JENKINS/Plugin+tutorial

verbruggenalex:
Thank you @myoung34, I'll give that a try.

fajran:
@verbruggenalex here's what I did

    $ git clone https://github.com/jenkinsci/workflow-multibranch-plugin.git
    $ cd workflow-multibranch-plugin
    $ git fetch origin pull/59/head:pr56
    $ git checkout pr56
    $ mvn package

After the build is done, you will find the plugin package file `target/workflow-multibranch.hpi` that you can upload to your Jenkins instance.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java
```diff
@@ -51,12 +51,17 @@
...
 
 /**
- * Checks out the desired version of {@link WorkflowBranchProjectFactory#SCRIPT}.
+ * Checks out the desired version of the script referred to by scriptPath.
  */
 class SCMBinder extends FlowDefinition {
 
     /** Kill switch for JENKINS-33273 in case of problems. */
     static /* not final */ boolean USE_HEAVYWEIGHT_CHECKOUT = Boolean.getBoolean(SCMBinder.class.getName() + ".USE_HEAVYWEIGHT_CHECKOUT"); // TODO 2.4+ use SystemProperties
+    private String scriptPath;
```
jglick: You need to either set this to `SCRIPT` in a `readResolve`, or make sure that all calls are prepared for it to be null. Otherwise you will be breaking existing branch projects after a plugin update.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactory.java
```diff
@@ -30,39 +30,54 @@
...
 import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.DataBoundSetter;
 
 /**
  * Recognizes and builds {@code Jenkinsfile}.
  */
 public class WorkflowBranchProjectFactory extends AbstractWorkflowBranchProjectFactory {
-    
     static final String SCRIPT = "Jenkinsfile";
+    private String scriptPath = SCRIPT;
```
jglick: Must also be restored in `readResolve`.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProjectFactory.java
```diff
@@ -29,17 +29,38 @@
...
+import org.apache.commons.lang.StringUtils;
 import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.DataBoundSetter;
+import java.io.IOException;
 
 /**
  * Defines organization folders by {@link WorkflowBranchProjectFactory}.
  */
 public class WorkflowMultiBranchProjectFactory extends AbstractWorkflowMultiBranchProjectFactory {
+    private String scriptPath = WorkflowBranchProjectFactory.SCRIPT;
```
jglick: `readResolve`

src/main/resources/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactory/config.jelly
```diff
@@ -24,6 +24,9 @@ THE SOFTWARE.
 -->
 
 <?jelly escape-by-default='true'?>
-<j:jelly xmlns:j="jelly:core">
+<j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
     <!-- TODO alternately, hide the selector, as in LiterateMultibranchProject/configure-factory.jelly -->
```
jglick: This comment is now obsolete and should be deleted.

src/test/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProjectTest.java
```diff
@@ -202,4 +202,50 @@ public DescriptorImpl() {
...
+        mp.setProjectFactory(projectFactory);
+
+        sampleRepo.write("file","initial commit");
+        sampleRepo.git("commit", "--all", "--message=init-master");
+
+        mp.scheduleBuild2(0).getFuture().get();
+        assertNull(mp.getItem("master"));
+
+        sampleRepo.git("checkout", "-b", "feature");
+        sampleRepo.write("another-Jenkinsfile", "echo \"branch=${env.BRANCH_NAME}\"; node {checkout scm; echo readFile('file')}");
```
jglick: BTW the `env.` syntax is now optional, so you could use just

```groovy
echo(/branch=$BRANCH_NAME/)
```

src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java
marvinthepa: fixed.

marvinthepa:
@jglick welcome back to this pull request.

Thanks for the helpful review comments (all implemented), I am learning a lot here.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java
```diff
@@ -59,8 +60,17 @@
     static /* not final */ boolean USE_HEAVYWEIGHT_CHECKOUT = Boolean.getBoolean(SCMBinder.class.getName() + ".USE_HEAVYWEIGHT_CHECKOUT"); // TODO 2.4+ use SystemProperties
     private String scriptPath;
 
+    public Object readResolve() {
+        this.scriptPath = WorkflowBranchProjectFactory.SCRIPT;
```
jglick: Uh, only if it was `null` before! Otherwise you will clobber a custom value on every restart!

src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java
```diff
@@ -59,8 +60,17 @@
...
+        return this;
+    }
+
     public SCMBinder(String scriptPath) {
-        this.scriptPath = scriptPath;
+        if (StringUtils.isEmpty(scriptPath)) {
```
jglick: This is not called from user input so it can (and should) assume that `scriptPath` is exact as passed.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactory.java
```diff
@@ -40,7 +40,12 @@
  */
 public class WorkflowBranchProjectFactory extends AbstractWorkflowBranchProjectFactory {
     static final String SCRIPT = "Jenkinsfile";
-    private String scriptPath = SCRIPT;
+    private String scriptPath;
+
+    public Object readResolve() {
+        this.scriptPath = WorkflowBranchProjectFactory.SCRIPT;
```
jglick: ditto, only if null

src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProjectFactory.java
```diff
@@ -38,7 +38,12 @@
  * Defines organization folders by {@link WorkflowBranchProjectFactory}.
  */
 public class WorkflowMultiBranchProjectFactory extends AbstractWorkflowMultiBranchProjectFactory {
-    private String scriptPath = WorkflowBranchProjectFactory.SCRIPT;
+    private String scriptPath;
+
+    public Object readResolve() {
+        this.scriptPath = WorkflowBranchProjectFactory.SCRIPT;
```
jglick: ditto, only if null

marvinthepa:
Sorry, stupid mistake(s).

I thought `readResolve` removes the need for initialization, which was stupid especially for tests.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java
```diff
@@ -58,19 +58,17 @@
 
     /** Kill switch for JENKINS-33273 in case of problems. */
     static /* not final */ boolean USE_HEAVYWEIGHT_CHECKOUT = Boolean.getBoolean(SCMBinder.class.getName() + ".USE_HEAVYWEIGHT_CHECKOUT"); // TODO 2.4+ use SystemProperties
-    private String scriptPath;
+    private String scriptPath = WorkflowBranchProjectFactory.SCRIPT;
```
jglick: BTW useless initializer—always overwritten by constructor.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactory.java
jglick: (done)

src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactory.java
```diff
@@ -30,39 +30,61 @@
...
+        if (this.scriptPath == null) {
+            this.scriptPath = WorkflowBranchProjectFactory.SCRIPT;
+        }
+        return this;
+    }
+
+    @DataBoundConstructor public WorkflowBranchProjectFactory() { }
+
+    @DataBoundSetter
+    public void setScriptPath(String scriptPath) {
```
jglick: Nice to have: `FormValidation DescriptorImpl.doCheckScriptPath(@QueryParameter String value)` to complain if you try to, say, use a `\` in the path.

jglick:
Seems to work fine, including with GitHub webhooks and multiple recognizers in an organization folder. Tested on https://github.com/cloudbeers/JENKINS-34561-demo using `jenkinsci/pipeline-as-code-github-demo:2017-06-01` amended with this snapshot.

crummy:
Thanks very much @marvinthepa and @jglick for all the work on this, been looking forward to this feature!

jonico:
@marvinthepa @jglick : First of all, thanks for this feature. Going through the code and our Jenkins instance hosted by CloudBees, I could not figure out how to make use of ```scriptPath``` with the default project recognizer in the GitHub org plugin.

The new parameter was only showing up in the custom script project recognizer where I would also have to specify the SCM:

image

Is there any possibility I missed to specify ```scriptPath``` for the default project recognizer or would I need to use the custom one, use ```Jenkinsfile``` as marker file, specify no SCM but the script path?

https://issues.jenkins-ci.org/browse/JENKINS-34561 which got fixed by this PR explicitly mentions ```Github organization folder``` in its problem description:

> In multibranch pipeline and Github organization folder the name Jenkinsfile is hard-coded and always looks for the top level folder.

marvinthepa:
@jonico I don't know, I didn't know about "default project recognizer" until I read it in your comment 😞 .

nickbroon:
Should one of options below work for adding multiple paths?

```
  projectFactories {
    workflowMultiBranchProjectFactory {
      scriptPath 'Jenkinsfile'
      scriptPath 'debian/Jenkinsfile'
    }
  }
```

```
  projectFactories {
    workflowMultiBranchProjectFactory {
      scriptPath 'Jenkinsfile'
    }
    workflowMultiBranchProjectFactory {
      scriptPath 'debian/Jenkinsfile'
    }
  }
```

jonico:
@nickbroon: In which file would one put your snippet? If it was part of a `Jenkinsfile` one would only be able to delegate the problem.

PR #62, opened 2017-05-15
make SCMBinder class public so we can reference it outside of this pa…

rawlingsj:
…ckage and plugin

We have a [plugin](https://github.com/fabric8io/jenkins-sync-plugin) that needs to check the job definitation before carrying out an action, i.e.

```
if (definition instanceof SCMBinder){...}
```  

This PR makes the class public so we can use the instanceof.


rawlingsj:
Sure, we have a plugin that synchronises openshift build configs back and forth with jenkins jobs.  There's a bit of logic that we use to check a flow definition type and to support multi branch jobs we use the `SCMBinder` class which is currently not a public class.

https://github.com/fabric8io/jenkins-sync-plugin/blob/job-to-bc/src/main/java/io/fabric8/jenkins/openshiftsync/BuildConfigToJobMapper.java#L63-L67

rawlingsj:
Closing this PR as we've changed our plugin to avoid use of `SCMBinder`

PR #65, opened 2017-06-13
[FIXED JENKINS-44848] Stop removing JobPropertys defined outside step

abayer:
[JENKINS-44848](https://issues.jenkins-ci.org/browse/JENKINS-44848)

There's some special logic for handling the case where the previous
run had the `properties` step run, but does not have the
`JobPropertyTrackerAction` on it - i.e., the first run of an existing
job with the `properties` step after upgrading. In that case, we still
use the old behavior of removing all existing `JobProperty`s. But in
all future runs, only `JobProperty`s defined via `properties` will be
removed by `properties`.

Also removed the warning messages about removing `JobProperty`s in
non-multibranch jobs, since, well, we don't do that any more once this
lands.

cc @reviewbybees


src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStep.java
```diff
@@ -87,7 +89,20 @@
         @SuppressWarnings("unchecked") // untypable
         @Override protected Void run() throws Exception {
             Job<?,?> job = build.getParent();
-            boolean isMultibranch = isMultibranch(job);
+
+            Run<?,?> previousRun = build.getPreviousCompletedBuild();
```
jglick: What happens if we are doing concurrent builds? Can things get out of whack?

And what about when all historical builds are deleted as part of some routine cleanup?

May make more sense to store the action on the `Job` rather than on the `Run`, not sure.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStep.java
```diff
@@ -87,7 +89,20 @@
...
+            JobPropertyTrackerAction previousAction = null;
+            boolean previousHadStep = false;
+            if (previousRun != null) {
+                previousAction = previousRun.getAction(JobPropertyTrackerAction.class);
+
+                // If the previous run did not have the tracker action, check to see if it ran the properties step. This
+                // is to deal with first run after this change is added.
+                if (previousRun instanceof WorkflowRun) {
+                    previousHadStep = new DepthFirstScanner().findFirstMatch(((WorkflowRun) previousRun).getExecution(),
```
jglick: `FlowExecutionOwner.Executable` → `FlowExecutionOwner.get[OrNull]`

src/main/resources/org/jenkinsci/plugins/workflow/multibranch/Messages.properties
```diff
@@ -1,7 +1,3 @@
 ReadTrustedStep._has_been_modified_in_an_untrusted_revis=\u2018{0}\u2019 has been modified in an untrusted revision
 WorkflowMultiBranchProject.DisplayName=Multibranch Pipeline
 WorkflowMultiBranchProject.Description=Creates a set of Pipeline projects according to detected branches in one SCM repository.
-JobPropertyStep._could_remove_warning=WARNING: The 'properties' step will remove all 'JobProperty's currently configured in this job, \
-  either from the UI or from an earlier 'properties' step.\n\
-  This includes configuration for discarding old builds, parameters, concurrent builds and build triggers.
-JobPropertyStep._removed_property_warning=WARNING: Removing existing job property ''{0}''
```
jglick: patch `help.html` to match please

src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyTrackerAction.java
```diff
@@ -0,0 +1,44 @@
...
+import javax.annotation.Nonnull;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Invisible action used for tracking what {@link JobProperty}s were defined in the Jenkinsfile for a given run.
+ */
+public class JobPropertyTrackerAction extends InvisibleAction {
```
jglick: Need not be `public`.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyTrackerAction.java
```diff
@@ -0,0 +1,44 @@
...
+    private final Set<String> jobPropertyDescriptors = new HashSet<>();
```
jglick: Javadoc link to `Descriptor.getId`

src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyTrackerAction.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public JobPropertyTrackerAction(@CheckForNull List<JobProperty> jobProperties) {
```
jglick: Why is this not `@Nonnull`?

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
```diff
@@ -230,6 +230,45 @@ public void resetStartsAndStops() {
         assertNull(b3.getPreviousBuild());
     }
 
+    @Issue("JENKINS-44848")
+    @Test public void onlyRemoveJenkinsfileProperties() throws Exception {
```
jglick: You also need a `@LocalData` test to demonstrate the behavior after an upgrade.

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
```diff
@@ -230,6 +230,45 @@ public void resetStartsAndStops() {
...
+        // Base case - not calling the properties step
+        p.setDefinition(new CpsFlowDefinition("echo 'Not doing anything'", true));
+
+        WorkflowRun b1 = r.buildAndAssertSuccess(p);
+
+        assertNotNull(p.getProperty(DisableConcurrentBuildsJobProperty.class));
+        assertNull(b1.getAction(JobPropertyTrackerAction.class));
+
+        // Adding a property, make sure the predefined one is still there.
+        p.setDefinition(new CpsFlowDefinition("properties([[$class: 'BuildDiscarderProperty', strategy: [$class: 'LogRotator', numToKeepStr: '1']]])", true));
```
jglick:
```java
// TODO Jenkins 2.x use symbols
```

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
```diff
@@ -230,6 +230,45 @@ public void resetStartsAndStops() {
...
+
+        WorkflowRun b2 = r.buildAndAssertSuccess(p);
+
+        assertNotNull(p.getProperty(DisableConcurrentBuildsJobProperty.class));
+        assertNotNull(p.getProperty(BuildDiscarderProperty.class));
+        JobPropertyTrackerAction action2 = b2.getAction(JobPropertyTrackerAction.class);
+        assertNotNull(action2);
+        assertEquals(1, action2.getJobPropertyDescriptors().size());
+        assertEquals(r.jenkins.getDescriptor(BuildDiscarderProperty.class).getId(),
+                action2.getJobPropertyDescriptors().iterator().next());
```
jglick:
```java
assertEquals(Collections.singleton(r.jenkins.getDescriptor(BuildDiscarderProperty.class).getId()), action2.getJobPropertyDescriptors());
```

src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStep.java
abayer: Hmm. Good question. Lemme think on that.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStep.java
abayer: *nod*

src/main/resources/org/jenkinsci/plugins/workflow/multibranch/Messages.properties
abayer: *nod*

src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyTrackerAction.java
abayer: *nod*

src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyTrackerAction.java
abayer: No good reason. Will fix.

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
abayer: Good point.

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
abayer: *nod*

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
abayer: *nod*

jglick:
> May make more sense to store the action on the `Job` rather than on the `Run`, not sure.

My reasoning is that we are dealing with `Job`-level configuration. So, we want to know which properties were added to the `Job` the last time this step was run. Which might have been in the last build, if there was one, but need not have been; pathologically

```groovy
sleep 10 - (BUILD_NUMBER as int)
properties([…])
```

Setting aside race conditions for the moment (which is already theoretically an issue with this step—probably it should acquire a per-`Job` lock during its execution, unless some of the add/remove/replace calls invoke foreign code), what you are interested in is merely the last-saved state.

Now when we are talking about an upgrade scenario, checking for a step execution in the last completed build is good enough—we are making a best effort here. But for routine ongoing usage, per-build metadata seems inherently unreliable.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStep.java
```diff
@@ -96,39 +111,37 @@
             }
             BulkChange bc = new BulkChange(job);
             try {
-                if (!isMultibranch) {
-                    l.getLogger().println(Messages.JobPropertyStep__could_remove_warning());
-                }
                 for (JobProperty prop : job.getAllProperties()) {
                     if (prop instanceof BranchJobProperty) {
                         // TODO do we need to define an API for other properties which should not be removed?
```
jglick: Not any more, so delete this comment. Ideally we could also delete the special-case handling of `BranchJobProperty`, but we probably need to keep it to handle some cases of upgraded branch projects: e.g., no remaining build history, or the first build after the upgrade is the one introducing `properties`. Unless you are confident that the logic below will never remove a property which was added via step.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStep.java
abayer: Switching to putting it on the `Job`.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStep.java
abayer: *nod*

abayer:
@jglick Ok, I think I got it with the action on the job - lemme know what you think.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStep.java
```diff
@@ -91,16 +93,21 @@
             Job<?,?> job = build.getParent();
 
             Run<?,?> previousRun = build.getPreviousCompletedBuild();
-            JobPropertyTrackerAction previousAction = null;
+            JobPropertyTrackerAction previousAction = job.getAction(JobPropertyTrackerAction.class);
             boolean previousHadStep = false;
-            if (previousRun != null) {
-                previousAction = previousRun.getAction(JobPropertyTrackerAction.class);
-
+            if (previousAction == null && previousRun != null) {
```
jglick: Nit: no need to call `getPreviousCompletedBuild` unless `previousAction == null`.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStep.java
```diff
@@ -113,7 +120,6 @@
             try {
                 for (JobProperty prop : job.getAllProperties()) {
                     if (prop instanceof BranchJobProperty) {
-                        // TODO do we need to define an API for other properties which should not be removed?
```
jglick: If you are deciding that there might be corner cases not fully covered by the main logic, best to leave a comment explaining that `BranchJobProperty` is considered especially critical, etc.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyTrackerAction.java
```diff
@@ -13,27 +13,19 @@
 /**
  * Invisible action used for tracking what {@link JobProperty}s were defined in the Jenkinsfile for a given run.
```
jglick: `for a given run` perhaps now misleading

src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyTrackerAction.java
jglick: (resolved)

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
```diff
@@ -264,11 +265,33 @@ public void resetStartsAndStops() {
...
+        WorkflowRun b1 = p.getLastBuild();
+        assertNotNull(b1);
+        assertNotNull(p.getProperty(BuildDiscarderProperty.class));
+        assertNull(p.getAction(JobPropertyTrackerAction.class));
+
+        p.setDefinition(new CpsFlowDefinition("properties([disableConcurrentBuilds()])", true));
+
+        r.buildAndAssertSuccess(p);
+
+        assertNull(p.getProperty(BuildDiscarderProperty.class));
```
jglick: So this is tracking the case that `b1` used `properties` to add `BuildDiscarderProperty` IIUC?

Do we need upgrade tests for the cases that

* the last build did not seem to run `properties`
* there are no (completed) builds at all

?

src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStep.java
```diff
@@ -87,7 +91,25 @@
         @SuppressWarnings("unchecked") // untypable
         @Override protected Void run() throws Exception {
             Job<?,?> job = build.getParent();
-            boolean isMultibranch = isMultibranch(job);
+
+            Run<?,?> previousRun = build.getPreviousCompletedBuild();
+            JobPropertyTrackerAction previousAction = job.getAction(JobPropertyTrackerAction.class);
+            boolean previousHadStep = false;
+            if (previousAction == null && previousRun != null) {
+                // If the previous run did not have the tracker action, check to see if it ran the properties step. This
```
jglick: Comment is now misleading.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStep.java
abayer: Oops. Fixing.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStep.java
abayer: Yup, fixing.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStep.java
abayer: *nod*

src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyTrackerAction.java
abayer: Fixing.

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
abayer: Correct - `properties` was called in `b1`.

I think I can tweak `onlyRemoveJenkinsfileProperties` to cover the cases you mention - if there's no `JobPropertyTrackerAction`, then there's no difference between a build without the `properties` step having been run and/or no completed builds.

jglick: True, that would be simpler.

abayer: Hmm. Clarifying - the no `JobPropertyTrackerAction` present but a `JobProperty` present case is covered in `onlyRemoveJenkinsfileProperties` - the first run has no `properties` step and so is not setting `JobPropertyTrackerAction` at all. I've cleaned up the test a bit to verify the absence of the action after that run. And I think all the other tests in `JobPropertyStepTest` end up covering the no-completed-builds scenario?

jglick: Whatever you say.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStep.java
```diff
@@ -92,20 +92,22 @@
...
-                        FlowExecution execution = owner.getOrNull();
-                        if (execution != null) {
-                            previousHadStep = new DepthFirstScanner().findFirstMatch(execution,
-                                    new NodeStepTypePredicate(step.getDescriptor())) != null;
+            if (previousAction == null) {
+                Run<?,?> previousRun = build.getPreviousCompletedBuild();
+                if (previousRun != null) {
+                    // If the job doesn't have the tracker action but does have a previous completed build,, check to
+                    // see if it ran the properties step. This is to deal with first run after this change is added.
+                    if (previousRun instanceof FlowExecutionOwner.Executable) {
```
jglick: null check and `instanceof` could be collapsed

src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyTrackerAction.java
jglick: (reverted)

src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStep.java
abayer: Whoops. An earlier iteration was more complex.

abayer:
Note that `JobPropertyTrackerAction` got switched back to being `public` - while I don't know right now exactly how it'll be used in Declarative, the intent is to leverage it down the road, so I'd rather have it be public now than have to deal with making it public later.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStep.java
```diff
@@ -96,18 +96,16 @@
             boolean previousHadStep = false;
             if (previousAction == null) {
                 Run<?,?> previousRun = build.getPreviousCompletedBuild();
-                if (previousRun != null) {
+                if (previousRun != null && previousRun instanceof FlowExecutionOwner.Executable) {
```
jglick: FWIW my point was that this is equivalent to the simpler

```java
if (previousRun instanceof FlowExecutionOwner.Executable) {
```

abayer: Well, jeez, I'm kinda dumb sometimes. =)

abayer: i.e., I always forget that `instanceof` will return false for a null object.

PR #68, opened 2017-12-12
Durability options

svanoort:
- [x] Add BranchProperty to let users specify a durability level for branches *before* they run.
- [x] Test the property step and BranchProperty works


pom.xml
```diff
@@ -28,12 +28,12 @@ THE SOFTWARE.
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.30</version>
+        <version>2.35</version>
```
jglick: 3.2 available.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/DurabilityHintBranchProperty.java
```diff
@@ -0,0 +1,82 @@
...
+                    @NonNull List<JobProperty<? super P>> jobProperties) {
+                List<JobProperty<? super P>> result = asArrayList(jobProperties);
+                for (Iterator<JobProperty<? super P>> iterator = result.iterator(); iterator.hasNext(); ) {
+                    JobProperty<? super P> p = iterator.next();
+                    if (p instanceof DurabilityHintJobProperty) {
+                        iterator.remove();
+                    }
+                }
+                if (getHint() != null) {
+                    result.add((JobProperty)(new DurabilityHintJobProperty(getHint())));
```
jglick: This speaks to the need for something like

```java
public interface FlowDurabilityHintProvider extends ExtensionPoint
    @CheckForNull FlowDurabilityHint suggestFor(@Nonnull FlowExecution x);
    static @Nonnull FlowDurabilityHint suggestedFor(@Nonnull FlowExecution x) {
        for (FlowDurabilityHintProvider p : ExtensionList.lookup(FlowDurabilityHintProvider.class)) {
            FlowDurabilityHint h = p.suggestFor(x);
            if (h != null) {
                return h;
            }
            return FlowDurabilityHint.MAX_SURVIVABILITY;
        }
    }
}
```

with implementations

* job property
* branch property
* (optionally) folder property
* global

pom.xml
```diff
@@ -63,14 +63,14 @@ THE SOFTWARE.
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins.version>1.642.3</jenkins.version>
-        <scm-api.version>2.0.7</scm-api.version>
+        <java.level>8</java.level>
+        <jenkins.version>2.62</jenkins.version>
```
jglick: Why not 2.73.3 BTW?

pom.xml
svanoort: I don't bump dependencies just because one *can.*

pom.xml
svanoort: Minimum core version that supports the features we need, so maximum number of users can apply it. 

If I'd caught it, I would have blocked the upstream update that mandates core 2.62 as well until after the release of this -- a trivial cleanup isn't worth forcing people to a much later LTS, unless we're doing it anyway.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/DurabilityHintBranchProperty.java
svanoort: A loose adaption of this is now applied (simplified somewhat).

src/main/resources/org/jenkinsci/plugins/workflow/multibranch/DurabilityHintBranchProperty/config.jelly
```diff
@@ -23,13 +23,14 @@ THE SOFTWARE.
...
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form">
-    <f:entry title="${%Custom Durability Setting}" field="hint">
+    <f:entry title="${%Custom Branch Speed/Durability Level}" field="hint">
         <j:set var="defaultHint" value="${descriptor.defaultDurabilityHint}"/>
         <select class="setting-input" name="hint">
             <j:forEach var="it" items="${descriptor.durabilityHintValues}">
-                <f:option value="${it.name()}" selected="${instance != null ? it.name()==instance[field].name() : it.name==defaultHint}">
-                    ${it.description}
-                </f:option>
+                <j:set var="optionBody" encode="false">${it.description}</j:set>
```
jglick: as elsewhere

src/main/resources/org/jenkinsci/plugins/workflow/multibranch/DurabilityHintBranchProperty/help.xml
```diff
@@ -0,0 +1,50 @@
...
+ ~ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ ~ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ ~ THE SOFTWARE.
+ -->
+<div>
+    <p>
+        This setting allows users to change the default durability mode for running Pipelines.  In most cases this is a
+        trade-off between performance and the ability for running pipelines to resume after unplanned Jenkins outages.
+
+        <br><br><strong>What does this do?</strong>
```
jglick: as elsewhere

pom.xml
```diff
@@ -260,6 +261,24 @@ THE SOFTWARE.
             <version>2.2</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.modules</groupId>
+            <artifactId>sshd</artifactId>
+            <version>2.3</version>
+            <scope>test</scope><!-- from git-server via workflow-cps-global-lib and otherwise unavailable during tests -->
```
jglick: Sigh.

Maybe it would be best to make `workflow-cps-global-lib` be just its original function—global libraries using `git-server`—and introduce a new plugin `pipeline-libs` with the newer external libraries stuff. `workflow-cps-global-lib` would then depend on `pipeline-libs` to pick up some common infrastructure, and could be deprecated (dropped from `workflow-aggregator`, etc.) and not used in tests of other plugins. People with `workflow-cps-global-lib` already installed would see no change in behavior, but new installations would not normally get the Git service.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/DurabilityHintBranchProperty.java
```diff
@@ -1,28 +1,34 @@
...
+import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
-import java.util.Iterator;
-import java.util.List;
+import java.util.Optional;
 
 /**
  * Branch property so we can define per-branch durability policies, i.e. so feature branches aren't built durably but master is.
  * Also lets us set the durability level before the pipeline has run (a step ahead of the "properties" step).
+ *
```
jglick:
```html
<p>
```

src/main/java/org/jenkinsci/plugins/workflow/multibranch/DurabilityHintBranchProperty.java
```diff
@@ -39,44 +45,63 @@ public DurabilityHintBranchProperty(@Nonnull FlowDurabilityHint hint) {
...
 
-        public static FlowDurabilityHint getDefaultHint() {
+        public static FlowDurabilityHint getDefaultDurabilityHint() {
             return GlobalDefaultFlowDurabilityLevel.getDefaultDurabilityHint();
         }
+
+        @Override
+        /** Lower ordinal than {@link org.jenkinsci.plugins.workflow.job.properties.DurabilityHintJobProperty} so those can override. */
+        public int ordinal() {
+            return 200;
```
jglick: Still makes no sense to me that this is not just

```java
@Extension(ordinal=200)
```

src/main/java/org/jenkinsci/plugins/workflow/multibranch/DurabilityHintBranchProperty.java
```diff
@@ -39,44 +45,63 @@ public DurabilityHintBranchProperty(@Nonnull FlowDurabilityHint hint) {
...
+        }
+
+        @CheckForNull
+        @Override
+        /**
+         * Dynamically fetch the property with each build, because the {@link BranchPropertyStrategy} does not re-evaluate,
```
jglick: I suppose the issue here is that `BranchProjectFactory.decorate` is called only when the branch project is created, not after the `MultiBranchProject` is reconfigured? If so, smells like a bug in `branch-api` to me. As mentioned in https://github.com/jenkinsci/workflow-job-plugin/pull/75, branch properties have been nearly unused so far.

Anyway fetching on the fly is fine (certainly prefer to modifying the job `config.xml` when that is avoidable).

src/main/java/org/jenkinsci/plugins/workflow/multibranch/DurabilityHintBranchProperty.java
```diff
@@ -39,44 +45,63 @@ public DurabilityHintBranchProperty(@Nonnull FlowDurabilityHint hint) {
...
+         * resulting in {@see <a href="https://issues.jenkins-ci.org/browse/JENKINS-48826">JENKINS-48826</a>}. */
+        public FlowDurabilityHint suggestFor(@Nonnull Item x) {
+            // BranchJobProperty *should* be present if it's a child of a MultiBranchProject but we double-check for safety
+            if (x instanceof WorkflowJob && x.getParent() instanceof MultiBranchProject && ((WorkflowJob)x).getProperty(BranchJobProperty.class) != null) {
+                MultiBranchProject mp  = (MultiBranchProject)(x.getParent());
+                WorkflowJob job = (WorkflowJob)x;
+                BranchJobProperty bjp = job.getProperty(BranchJobProperty.class);
+
+                String sourceId = bjp.getBranch().getSourceId();
+                if (sourceId != null) {
```
jglick: Huh, surprised that this is potentially nullable. I think in practice it is not, but better to be defensive until the annotations are cleaned up upstream.

src/test/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProjectTest.java
```diff
@@ -188,12 +197,22 @@ public DescriptorImpl() {
...
+        r.waitForCompletion(p.getLastBuild());
+        Thread.sleep(1000);
+        assert !p.isBuilding();
         WorkflowRun b1 = p.getLastBuild();
         assertEquals(1, b1.getNumber());
-        mp.scheduleBuild2(0).getFuture().get();
-        mp.getIndexing().writeWholeLogTo(System.out);
-        assertEquals("[p, p/master]", ExtensionList.lookup(Listener.class).get(0).names.toString());
+        Queue.Item it = mp.scheduleBuild2(0);
+        Thread.sleep(100);
```
jglick: Beware—this smells like something that could flake out on a loaded CI builder.

PR #69, opened 2017-12-18
Make SCMBinder try to abort builds in which Jenkinsfile has been modified by an untrusted contributor

jglick:
https://github.com/jenkinsci/ssh-slaves-plugin/pull/79 and the like are pretty annoying: the historical behavior of `SCMBinder` is to allow a build which includes an untrusted edit to `Jenkinsfile` to proceed, but actually reading the `Jenkinsfile` contents from the corresponding trusted revision instead. If you get a green check mark, no one is going to go looking for the one-line warning about this. It is better to behave as if the script started with

```groovy
readTrusted 'Jenkinsfile'
```

In other words, to just fail immediately in such a case. I cannot think of any good reason for anyone to be relying on the old behavior, but just in case I kept it as an option which can be enabled by system property.

Note that this change will not have any effect on GitHub PR merge jobs until [JENKINS-43194](https://issues.jenkins-ci.org/browse/JENKINS-43194) is implemented. Performing a rev-to-tip comparison of script contents in a heavyweight checkout would be pretty awkward, as `SCMBinder` could not then rely on `CpsScmFlowDefinition` for script checkout; `ReadTrustedStep` would need to be refactored to make its main logic reusable, and there would be a lot of different code paths to maintain code coverage for. I think the effort is better put into JENKINS-43194.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java
@@ -24,6 +24,7 @@
 
 package org.jenkinsci.plugins.workflow.multibranch;
 
+import hudson.AbortException;
 import hudson.Extension;
 import hudson.Functions;
 import hudson.model.Action;
@@ -58,6 +59,10 @@ class SCMBinder extends FlowDefinition {
 
     /** Kill switch for JENKINS-33273 in case of problems. */
     static /* not final */ boolean USE_HEAVYWEIGHT_CHECKOUT = Boolean.getBoolean(SCMBinder.class.getName() + ".USE_HEAVYWEIGHT_CHECKOUT"); // TODO 2.4+ use SystemProperties
+
+    /** Kill switch for making this as strict as {@link ReadTrustedStep} about untrusted modifications. */
+    static /* not final */ boolean IGNORE_UNTRUSTED_EDITS = Boolean.getBoolean(SCMBinder.class.getName() + ".IGNORE_UNTRUSTED_EDITS"); // TODO 2.4+ use SystemProperties
+
     private String scriptPath = WorkflowBranchProjectFactory.SCRIPT;
 
     public Object readResolve() {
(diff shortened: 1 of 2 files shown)


svanoort:
Agree with @abayer that this needs some extra consideration for unexpected impacts

rsandell:
I don't have the context to see any potential consequences on this.

Another mode that would be nice would be if Jenkins could pause and ask something like

"__WARNING__ The script has been modified by a non trusted party, please review before proceeding. Proceed? [Yes](#yes) / [No](#no)"

jglick:
@rsandell that would be a reasonable enhancement, although the link could only safely be clicked by someone with `Item.CONFIGURE`, so it would not for example help the case on ci.jenkins.io where few developers have anything beyond r/o permissions on Jenkins, even when they *do* have write access to the repository backing a job. Also note that the link would not survive Jenkins restart (the program has not started at this point, and you are blocking a native thread until someone interacts with the build).

Anyway that is something that could be added later—this PR at least fixes the current confusing behavior, and does so with minimal code changes.

jglick:
> people not expecting their PR build to fail because it also happened to contain a change to the `Jenkinsfile`

My attitude is that this is better than silently ignoring the change. You can always request that a maintainer refile the same branch. Or if [JENKINS-46795](https://issues.jenkins-ci.org/browse/JENKINS-46795) were implemented, it would suffice for the maintainer to approve the PR and kick off a new build.

abayer:
@jglick Yeah, I think I might prefer loud logging in these cases rather than aborting, as mentioned in JENKINS-45970.

jglick:
Well decide what you want one way or another and the PR can be adjusted accordingly. The worst thing you can do is leave things the way they are.

abayer:
@jglick Yeah, let's go with logging a loud warning, then. =)

jglick:
OK. I still think a simple build failure is a lot more discoverable than any kind of build logging, but will do whatever is necessary to get something merged.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java
```diff
@@ -141,4 +164,21 @@ public SCMBinder(String scriptPath) {
 
     }
 
+    // TODO seems there is no general-purpose ConsoleNote which simply wraps markup in specified HTML
+    @SuppressWarnings("rawtypes")
+    public static class WarningNote extends ConsoleNote {
+
+        @Override public ConsoleAnnotator annotate(Object context, MarkupText text, int charPos) {
+            text.addMarkup(0, text.length(), "<span class='warning-inline'>", "</span>");
```
jglick: Thus:

screenshot

jglick:
Would add @dwnusbaum as a reviewer if I had write permission.

abayer:
Huh, I don't have write access either. That's...wrong.

abayer:
And fixed - the developers group didn't have write access for some reason.

PR #70, opened 2018-01-09
Updates

jglick:
Due to some behavioral incompatibility by @stephenc, trying to do some of these piecemeal causes `WorkflowMultiBranchProjectTest.conflictingBranches` to hang—branch indexing goes to sleep and does not wake up. Seems to be necessary to update at least `scm-api`.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -256,7 +256,7 @@ THE SOFTWARE.
         <dependency>
             <groupId>org.jenkins-ci.modules</groupId>
             <artifactId>sshd</artifactId>
-            <version>1.11</version>
+            <version>2.3</version>
             <scope>test</scope><!-- from git-server via workflow-cps-global-lib and otherwise unavailable during tests -->
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@ THE SOFTWARE.
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.30</version>
+        <version>3.2</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -63,14 +63,15 @@ THE SOFTWARE.
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins.version>1.642.3</jenkins.version>
-        <scm-api.version>2.0.7</scm-api.version>
+        <jenkins.version>2.60.3</jenkins.version>
+        <java.level>8</java.level>
+        <scm-api.version>2.2.6</scm-api.version>
         <no-test-jar>false</no-test-jar>
         <workflow-step-api.version>2.10</workflow-step-api.version>
         <workflow-scm-step-plugin.version>2.4</workflow-scm-step-plugin.version>
         <workflow-support.version>2.14</workflow-support.version>
         <workflow-cps.version>2.32</workflow-cps.version>
-
+        <git-plugin.version>3.7.0</git-plugin.version>
     </properties>
     <dependencies>
         <dependency>
@@ -91,7 +92,7 @@ THE SOFTWARE.
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>structs</artifactId>
-            <version>1.7</version>
+            <version>1.10</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -106,12 +107,18 @@ THE SOFTWARE.
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>branch-api</artifactId>
-            <version>2.0.7</version>
+            <version>2.0.17</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.jenkins-ci</groupId>
+                    <artifactId>annotation-indexer</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>cloudbees-folder</artifactId>
-            <version>5.18</version>
+            <version>6.1.2</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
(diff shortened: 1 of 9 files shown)


pom.xml
```diff
@@ -63,14 +63,15 @@ THE SOFTWARE.
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins.version>1.642.3</jenkins.version>
-        <scm-api.version>2.0.7</scm-api.version>
+        <jenkins.version>2.60.3</jenkins.version>
```
jglick: Avoidable with more work, but [who cares anyway](stats.jenkins.io/pluginversions/workflow-multibranch.html)?

pom.xml
```diff
@@ -106,12 +107,18 @@ THE SOFTWARE.
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>branch-api</artifactId>
-            <version>2.0.7</version>
+            <version>2.0.17</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.jenkins-ci</groupId>
+                    <artifactId>annotation-indexer</artifactId>
```
jglick: Needs to be fixed upstream I guess, or in a BOM.

pom.xml
```diff
@@ -106,12 +107,18 @@ THE SOFTWARE.
...
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>cloudbees-folder</artifactId>
-            <version>5.18</version>
+            <version>6.1.2</version>
```
jglick: Originally had 6.3, which needed a newer core, went back, then updated the core anyway…so could be 6.3. But should not matter.

pom.xml
```diff
@@ -208,14 +219,18 @@ THE SOFTWARE.
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>git</artifactId>
-            <version>3.0.4</version>
+            <version>${git-plugin.version}</version>
```
jglick: DRY

pom.xml
```diff
@@ -241,19 +256,25 @@ THE SOFTWARE.
         <dependency>
             <groupId>org.jenkins-ci.modules</groupId>
             <artifactId>sshd</artifactId>
-            <version>1.6</version>
+            <version>2.3</version>
```
jglick: tricky

pom.xml
```diff
@@ -241,19 +256,25 @@ THE SOFTWARE.
...
             <scope>test</scope><!-- from git-server via workflow-cps-global-lib and otherwise unavailable during tests -->
         </dependency>
         <dependency>
-            <groupId>org.jenkins-ci.plugins</groupId>
-            <artifactId>matrix-auth</artifactId>
```
jglick: was gratuitous

pom.xml
```diff
@@ -241,19 +256,25 @@ THE SOFTWARE.
...
+            <groupId>org.jenkins-ci.modules</groupId>
+            <artifactId>instance-identity</artifactId>
+            <version>2.1</version>
```
jglick: transitive dep of `sshd`

pom.xml
```diff
@@ -241,19 +256,25 @@ THE SOFTWARE.
...
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.modules</groupId>
+            <artifactId>ssh-cli-auth</artifactId>
```
jglick: ditto

pom.xml
```diff
@@ -241,19 +256,25 @@ THE SOFTWARE.
...
             <version>1.4</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>credentials</artifactId>
-            <version>2.1.11</version>
+            <version>2.1.14</version>
```
jglick: `requireUpperBoundDeps`

src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStep.java
```diff
@@ -79,7 +78,7 @@
     }
 
     public Map<JobPropertyDescriptor,JobProperty> getPropertiesMap() {
-        return Descriptor.toMap(properties);
+        return Descriptor.toMap((List) properties);
```
jglick: `-source 8`

src/main/java/org/jenkinsci/plugins/workflow/multibranch/ReadTrustedStep.java
```diff
@@ -170,7 +170,7 @@ public String getPath() {
                 if (tip == null) {
                     throw new AbortException("Could not determine exact tip revision of " + branch.getName());
                 }
-                build.addAction(new SCMRevisionAction(tip));
+                build.addAction(new SCMRevisionAction(scmSource, tip));
```
jglick: https://github.com/jenkinsci/scm-api-plugin/pull/37#discussion_r119865176 :sigh:

src/test/java/org/jenkinsci/plugins/workflow/multibranch/ReplayActionTest.java
```diff
@@ -120,18 +115,13 @@
...
-        perms.put(Item.CONFIGURE, Collections.singleton("dev1")); // implies REPLAY
-        perms.put(ReplayAction.REPLAY, Collections.singleton("dev2"));
-        perms.put(Item.BUILD, Collections.singleton("dev3")); // does not imply REPLAY
-        top.addProperty(new AuthorizationMatrixProperty(perms));
+        r.jenkins.setAuthorizationStrategy(new MockAuthorizationStrategy().
+            grant(Jenkins.ADMINISTER).everywhere().to("admin").
+            grant(Jenkins.READ).everywhere().to("dev1", "dev2", "dev3").
+            grant(Item.CONFIGURE).onFolders(top).to("dev1"). // implies REPLAY
+            grant(ReplayAction.REPLAY).onFolders(top).to("dev2").
+            grant(Item.BUILD).onFolders(top).to("dev3")); // does not imply REPLAY
```
jglick: More readable I hope.

pom.xml
svanoort: Yeah, we are going to be mandating 2.60.x here-on-out for Java 8 since the upstream plugins depend on it now anyway (and durability work means we'll need to bump to Java 8 versions).

src/main/java/org/jenkinsci/plugins/workflow/multibranch/ReadTrustedStep.java
svanoort: Adding an extra :sigh: about that for good measure.

jglick:
Found a bug in `branch-api`; working on an apparent regression in `ResolveScmStep`.

jglick:
@reviewbybees done

PR #72, opened 2018-03-11
[JENKINS-26143] Tests

daniel-beck:
Companion PR to core change https://github.com/jenkinsci/jenkins/pull/3014

Built against https://github.com/jenkinsci/jenkins/pull/3014/commits/7b8e4b1e093550dc8ca1f3664ed161e9c24a93db

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,8 +64,8 @@ THE SOFTWARE.
     </pluginRepositories>
     <properties>
         <java.level>8</java.level>
-        <jenkins-core.version>2.111-20180311.231847-1</jenkins-core.version>
-        <jenkins-war.version>2.111-20180311.232051-1</jenkins-war.version>
+        <jenkins-core.version>2.111-20180312.014148-2</jenkins-core.version>
+        <jenkins-war.version>2.111-20180312.014313-2</jenkins-war.version>
         <scm-api.version>2.2.6</scm-api.version>
         <no-test-jar>false</no-test-jar>
         <workflow-step-api.version>2.13</workflow-step-api.version>
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,8 @@ THE SOFTWARE.
     </pluginRepositories>
     <properties>
         <java.level>8</java.level>
-        <jenkins.version>2.62</jenkins.version>
+        <jenkins-core.version>2.111-20180311.231847-1</jenkins-core.version>
+        <jenkins-war.version>2.111-20180311.232051-1</jenkins-war.version>
         <scm-api.version>2.2.6</scm-api.version>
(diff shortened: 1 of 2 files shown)


daniel-beck:
Failing test looks unrelated, probably a slightly outdated core that doesn't have https://github.com/jenkinsci/jenkins/pull/3342 yet.

pom.xml
```diff
@@ -64,7 +64,8 @@ THE SOFTWARE.
     </pluginRepositories>
     <properties>
         <java.level>8</java.level>
-        <jenkins.version>2.62</jenkins.version>
+        <jenkins-core.version>2.111-20180312.014148-2</jenkins-core.version>
+        <jenkins-war.version>2.111-20180312.014313-2</jenkins-war.version>
```
jglick: You should also delete the now-older baseline from `Jenkinsfile`, thus simply:

```groovy
buildPlugin()
```

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
```diff
@@ -99,6 +101,7 @@
         ParametersDefinitionProperty.DescriptorImpl.class.isAnnotationPresent(Symbol.class) && // "parameters"
         BooleanParameterDefinition.DescriptorImpl.class.isAnnotationPresent(Symbol.class) && // "booleanParam"
         StringParameterDefinition.DescriptorImpl.class.isAnnotationPresent(Symbol.class) && // "string"
+        ChoiceParameterDefinition.DescriptorImpl.class.isAnnotationPresent(Symbol.class) && // "choice" and "choiceParam"
```
jglick: Well clearly if we have this new core baseline, `HAVE_SYMBOL == true`, so we can just delete all these switches.

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
```diff
@@ -206,6 +211,44 @@ public void resetStartsAndStops() {
...
+            r.assertLogContains("value bar", b9);
+        }
+
+    }
+
+    @Issue("JENKINS-26143")
+    @Test
+    public void testChoiceParameterSnippetizer() throws Exception {
+        //new SnippetizerTester(r).assertGenerateSnippet();
+        new SnippetizerTester(r).assertRoundTrip(new JobPropertyStep(Arrays.asList(new ParametersDefinitionProperty(new ChoiceParameterDefinition[] {
```
jglick: use varargs

jglick:
Reminder that that does _not_ mean this needs to be closed. Merging #81 would automatically mark this as merged as well.

PR #75, opened 2018-05-17
Test and fix for [JENKINS-51290] - properties step fails with error about job not yet started if could not obtain execution for previous run

svanoort:
... unfortunately we may legitimately null the execution for certain critical failures.

[https://issues.jenkins-ci.org/browse/JENKINS-51290](https://issues.jenkins-ci.org/browse/JENKINS-51290)

@reviewbybees 

Note: A *better* fix would be to add a new API to `FlowExecutionOwner` that can trigger lazy loading and behaves like either `FlowExcecutionOwner#get()` or `FlowExcecutionOwner#getOrNull()` but this issue is fairly critical so we need at least a fast solution.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
@@ -145,7 +145,8 @@ public class JobPropertyStepTest {
     public void testPreviousBuildFailedHard() throws Exception {
         WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
 
-        // First we simulate a build that has failed so hard it has a null execution
+        // First we simulate a build that has resulted in a null execution
+        // This can be a result of a hard-kill of the Pipeline or a catastrophic failure starting the run, i.e. fetching the Jenkinsfile
         p.setDefinition(new CpsFlowDefinition("echo 'Not doing anything'", true));
         WorkflowRun run = r.buildAndAssertSuccess(p);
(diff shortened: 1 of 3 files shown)


src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
```diff
@@ -138,6 +140,24 @@ public void resetStartsAndStops() {
...
+    @Test
+    public void testPreviousBuildFailedHard() throws Exception {
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+
+        // First we simulate a build that has failed so hard it has a null execution
+        p.setDefinition(new CpsFlowDefinition("echo 'Not doing anything'", true));
+        WorkflowRun run = r.buildAndAssertSuccess(p);
+        Field f = run.getClass().getDeclaredField("execution");
+        f.setAccessible(true);
+        f.set(run, null);
```
rsandell: Doesn't look very close to a "real world scenario" to me?

svanoort: Yeah, It's not.  The real world scenario generate the same null execution, but requires a hard-kill of the Pipeline or a failure within the CPS VM thread.

oleg-nenashev: You could easily emulate it by a SCMCpsFlowDefinition pointing to a repo without Pipeline. FileSystem SCM (or small copy-paste of it) allows doing it without actual repos

PR #76, opened 2018-06-06
Update unit tests for compatibility with git-plugin >=3.9.x

kshultzCB:
# Description

This pull request updates unit tests in workflow-multibranch so that they are compatible with versions of git-plugin >=3.9.0. These changes to git-plugin went in with [PR 577](https://github.com/jenkinsci/git-plugin/pull/577). As such, there is now a test-scope maven dependency for git-plugin 3.9.1. `git-plugin.version` has been left at 3.7.0.

@svanoort and @rsandell , your reviews would be appreciated. Thanks!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -71,7 +71,7 @@ THE SOFTWARE.
         <workflow-scm-step-plugin.version>2.4</workflow-scm-step-plugin.version>
         <workflow-support.version>2.17</workflow-support.version>
         <workflow-cps.version>2.53</workflow-cps.version>
-        <git-plugin.version>3.7.0</git-plugin.version>
+        <git-plugin.version>3.9.1</git-plugin.version>
     </properties>
     <dependencies>
         <dependency>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -197,7 +197,7 @@ THE SOFTWARE.
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>git</artifactId>
-            <version>${git-plugin.version}</version>
+            <version>3.9.1</version>
```
abayer: You should be updating `git-plugin.version` in `properties` rather than just changing it here, fwiw.

kshultzCB: Roger that. Done.

jglick:
> a link to the PCT run that shows it passing

I think that is not necessary in general. The point of PCT is to alert you that there would be test failures were you to update test dependencies. Once you actually do that update and correct the failures, we can assume that it worked. If not, it will be another round.

svanoort:
@jglick It's perfectly reasonable to ask for a before-and-after as a matter of verification for changes

PR #77, opened 2018-08-28
Making the BranchJobProperty constructor public

mohsen0:
I would appreciate if you make this public, 
I am working on a different plugin which provides a different kind of project recognizer, I appreciate if this class can be instantiated outside of your package.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/multibranch/BranchJobProperty.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/multibranch/BranchJobProperty.java
@@ -47,7 +47,7 @@ public class BranchJobProperty extends WorkflowJobProperty {
 
     private @Nonnull Branch branch;
 
-    BranchJobProperty(@Nonnull Branch branch) {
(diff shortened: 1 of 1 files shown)


jglick:
> a different kind of project recognizer

Why would this require `BranchJobProperty` to be `public`? There are already separate plugins depending on this one which successfully implement different kinds of project recognizers with the existing APIs; the `BranchJobProperty` constructor is intended to be called only from `AbstractWorkflowBranchProjectFactory.setBranch`. Is source code for this plugin available somewhere?

mohsen0:
It is not open sourced yet, the problem I have is 
1) SCMVar requires you to use org.jenkinsci.plugins.workflow.multibranch.BranchJobProperty, but that class has a package-private constructor.
2) It mandates that the job parent is a WorkflowMultiBranchProject, which is not suitable for extending.

The plugin I am working on instead of different jenkinsfiles, for example if a config-ci.yml file exists, it will recognise the project and set the build script to a groovy script loaded as a library.

PR #78, opened 2018-09-06
Added public modifier to SCMBinder class

cnserd:
I am developing a plugin and I would like to identify jobs by their FlowDefinition types. I need SCMBinder to be public for this purpose.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java
@@ -54,7 +54,7 @@ import org.jenkinsci.plugins.workflow.job.WorkflowRun;
 /**
  * Checks out the desired version of the script referred to by scriptPath.
  */
-class SCMBinder extends FlowDefinition {
+public class SCMBinder extends FlowDefinition {
 
     /** Kill switch for JENKINS-33273 in case of problems. */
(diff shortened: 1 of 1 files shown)


waveywaves:
I am using SCMBinder for development purposes as well. Any idea when it would be possible to get this merged ?

cnserd:
In my case, I have access to the build as a WorkflowRun instance through a RunListener. And I would like to know if this pipeline is pulling code from version control. So I am checking if its FlowDefinition is of type:

- CpsFlowDefinition (in which case I check for BuildData),
- CpsScmFlowDefinition (in which case I read sources from GitSCM), or
- SCMBinder (in which case I look at the BranchJobProperty and then GitSCM)

except I can't access SCMBinder so as a workaround I had to hardcode its descriptor's display name which isn't ideal.

waveywaves:
@jglick I am adding a feature to https://github.com/openshift/jenkins-sync-plugin to do openshift builds through multibranch pipelines. Here is the PR for that, for now we have cloned the class to use it's functionality. https://github.com/openshift/jenkins-sync-plugin/pull/310

bitwiseman:
@jglick  
It sounds like you are strongly opposed to this change.  Is there something the contributor could do to change this or should this PR be closed?

jglick:
Perhaps not “strongly” but -0 since this is effectively introducing a new API without a clear use case that I can see.

PR #83, opened 2019-02-28
[JENKINS-55116] Hack patch fix

rsandell:
Since we haven't been able yet to understand the root cause of
why the job configurations loose content at times this hack
should help to leviate the symtoms and allow the users to
do a rescan to repopulate the rest of the content in the config.

- [x] Needs more tests
- [x] Possibly more logging


rsandell:
So where I'm putting the code isn't well placed. During startup this code path is called waaaay too early (before WorkFlowJob#onLoad) so not much data to go on, the mixIn to retrieve the builds isn't created yet for example which is vital.
So I need to find another place to put this, but if I do it any later it will most likely already been filtered out by the ComputedFolder #sadface

PR #84, opened 2019-07-21
Consistently use sandbox for CpsFlowDefinitions

basil:
While looking at the test suite for this plugin, I noticed the `CpsFlowDefinition`s in the tests don't consistently use the script security sandbox. Using the script security sandbox results in a more realistic environment given that the script security sandbox should always be enabled in production.

In this change, I replaced any usages of the deprecated single-argument constructor for `CpsFlowDefinition` with usages of the non-deprecated two-argument constructor, passing in `true` as the second argument in order to always enable the script security sandbox.

Note that I did _not_ touch `ReplayActionTest#multibranch`, which _intentionally_ disables the script security sandbox for testing purposes.


dwnusbaum:
@basil The GitHub UI kind of bugged out on me when I went to merge this, and now it says its conflicted. I'm not really sure what happened, but I don't think there should be any conflicts. I'll try to close and reopen to see if that does anything.

dwnusbaum:
Weird, it did get merged, see 488f87fa6af4925cb5a7cb60aa365d1c6b36f60c, looks like some kind of concurrency issue.

PR #87, opened 2019-08-09
Trigger Job Feature

aytuncbeken:
Hi,
I wanted to add feature to MultiBranch Pipeline where triggering other jobs is possible when a Pipeline is created or deleted by Branch Indexing. This can be solution to similar issues like [this](https://issues.jenkins-ci.org/browse/JENKINS-40606). I Hope this fits with your plugin goals.

Thanks.


aytuncbeken:
@rsandell I will be glad if you can respond. If this change is not fitting your pipeline goals, I can start a new one. 
Thanks a lot.

PR #88, opened 2019-08-16
[JENKINS-58965] Deliver enable/disable job feature from UI

pjanouse:
See [JENKINS-58965](https://issues.jenkins-ci.org/browse/JENKINS-58965).

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProject.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProject.java
@@ -134,6 +134,12 @@ public class WorkflowMultiBranchProject extends MultiBranchProject<WorkflowJob,W
                 strategy != null ? strategy.getPropertiesFor(head) : Collections.<BranchProperty>emptyList()));
     }
 
+    // Expose folders disabling feature - acts only as a proxy here because parent is a protected method
+    @Override
+    public boolean supportsMakeDisabled() {
+        return super.supportsMakeDisabled();
+    }
+
     @Extension public static class DescriptorImpl extends MultiBranchProjectDescriptor implements IconSpec {
 
         @Override public String getDisplayName() {


MaciejKucia:
That feature would be awesome. Thank you for this PR!

src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProject.java
```diff
@@ -134,6 +134,12 @@ private BranchJobProperty reconstructBranchJobProperty(@Nonnull WorkflowJob job,
                 strategy != null ? strategy.getPropertiesFor(head) : Collections.<BranchProperty>emptyList()));
     }
 
+    // Expose folders disabling feature - acts only as a proxy here because parent is a protected method
```
dwnusbaum: This comment doesn't make sense to me. It looks like the inheritance hierarchy here is `WorkflowMultiBranchProject` -> `MultiBranchProject` -> `ComputedFolder`. Adding an override that just calls the existing super method shouldn't change the behavior unless code is checking the access modifier of the method somewhere, which would be strange. It looks like the [super implementation from ComputedFolder](https://github.com/jenkinsci/cloudbees-folder-plugin/blob/f7b0a54b3b4d6b267dde1f18662f2227b376ca8b/src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java#L391) (MultiBranchProject doesn't implement it as far as I can tell) already returns true, so I would think that either these projects can already be disabled from the UI without this PR, or that something else is making it not possible and this PR wouldn't help. 

Did you test the plugin with and without this PR to make sure it fixed the issue?

pjanouse: As far as I can tell you are almost right. The root cause is Jelly (unlike e.g. script in script console) can't call a protected method (verified). That's the reason why an exception occurred (cannot call, method not found or sth similar, which is swallowed) and in that case it is taken as false/otherwise case during UI generation and the link isn't generated. That's the reason the fix consists simply of an overriding the visibility of the method and not the impl. (used parent impl. whatever it is).

I tried my best to explain what was wrong, how I tried to fix it and the reason why it is important for us in the description of JENKINS-58965.

Finally an answer to your question - yes, I've tested both cases thoroughly before I created this PR, I'm confident it can help some users in our community.

dwnusbaum: Ah ok that makes sense. It looks like the visibility on those methods was changed to `public` in cloudbees-folder 6.10.0 after you filed this PR, see https://github.com/jenkinsci/cloudbees-folder-plugin/pull/143, so maybe everything already works even without this patch when running a recent version of cloudbees-folder?

dwnusbaum: Ok, I just confirmed that the "Disable Multibranch Pipeline" button shows up correctly when running cloudbees-folder 6.12 or newer. I will go ahead and file a PR to update all of the dependencies here, but I think we can close this PR. Thanks for submitting it!

pjanouse: Yes, there is almost a year when I made a fix for an issues reported by our users, happy to hear it is fixed by more convenient way. 
You are right we should rely on updated (fixed) version of Folder plugin and I'm going to close this PR.

PR #90, opened 2019-12-17
Bump parent pom to 3.54

slonopotamus:


pom.xml
```diff
@@ -204,6 +204,13 @@ THE SOFTWARE.
             <version>2.7.1</version>
             <scope>test</scope>
         </dependency>
+        <!-- TODO: Remove this when https://github.com/jenkinsci/subversion-plugin/pull/240 is fixed -->
```
jglick: https://github.com/jenkinsci/subversion-plugin/pull/240 for the record.

This would better be placed in `dependencyManagement`.

I think it is time to consider just removing the `subversion` test dep here, and moving the affected test cases into the `subversion` plugin.

slonopotamus: AFAIK, to see the difference between `dependencies` and `dependencyManagement`, one needs to have a `pom.xml` that inherits from current one. From `dependencyManagement` documentation:

> Default dependency information for projects that inherit from this one. The dependencies in this section are not immediately resolved. Instead, when a POM derived from this one declares a dependency described by a matching groupId and artifactId, the version and other values from this section are used for that dependency if they were not already specified.

Do you actually expect someone to inherit from this `pom.xml`?

jglick: > to see the difference between dependencies and dependencyManagement, one needs to have a pom.xml that inherits from current one

No, the documentation is incomplete then. Moving this declaration to `dependencyManagement` would emphasize that we are not _necessarily_ adding such a dependency, but if we _did_, then it should be of the following version. This is recommended for all cases when a transitive dependency must be tied to a specific version but we have no independent reason for having that dependency at all.

> There are several tests in subversion that were also affected by this issue

I should have rephrased:

I think it is time to consider just removing the `subversion` test dep here, and moving all test cases which ran Subversion checkouts from this plugin into the subversion plugin.

slonopotamus: > This is recommended for all cases when a transitive dependency must be tied to a specific version but we have no independent reason for having that dependency at all.

OK, done.

jglick: > do I need to specify `<scope>test</scope>` or not?

Can never recall, but better to stay on the safe side and do so.

pom.xml
```diff
@@ -284,6 +284,17 @@ THE SOFTWARE.
             <scope>test</scope>
         </dependency>
     </dependencies>
+    <dependencyManagement>
+        <dependencies>
+            <!-- TODO: Remove this when https://github.com/jenkinsci/subversion-plugin/pull/240 is fixed -->
```
jglick:
```suggestion
            <!-- TODO: Remove this when https://github.com/jenkinsci/subversion-plugin/pull/240 is released -->
```

which I hope to do today. Care to review https://github.com/jenkinsci/subversion-plugin/pull/241?

pom.xml
slonopotamus: Ok, that's what I actually did :)

pom.xml
slonopotamus: > I hope to do today

Well, if that is going to happen so soon, we can completely avoid all these dances with workaround and _just wait_.

slonopotamus: Anyway, applied your suggestion.

pom.xml
```diff
@@ -284,6 +284,17 @@ THE SOFTWARE.
...
+        <dependencies>
+            <!-- TODO: Remove this when https://github.com/jenkinsci/subversion-plugin/pull/240 is released -->
+            <dependency>
+                <groupId>org.jenkins-ci.plugins</groupId>
+                <artifactId>mapdb-api</artifactId>
+                <version>1.0.9.0</version>
+                <scope>test</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
```
jglick: Try just updating the `subversion` dep to `2.13.0`.

slonopotamus:
Closing this in favor of #91.

pom.xml
slonopotamus: jenkinsci/workflow-multibranch-plugin#91

PR #91, opened 2019-12-18
Bump parent pom to 3.54 [v2]

slonopotamus:
This PR is a v2 of #90. It bumps subversion-plugin to 2.13.0 instead of adding a workaround for jenkinsci/subversion-plugin#240

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@ THE SOFTWARE.
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.25</version>
+        <version>3.54</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -66,7 +66,7 @@ THE SOFTWARE.
         <revision>2.22</revision>
         <changelist>-SNAPSHOT</changelist>
         <java.level>8</java.level>
-        <jenkins.version>2.121.1</jenkins.version>
+        <jenkins.version>2.138.4</jenkins.version>
         <scm-api.version>2.2.7</scm-api.version>
         <no-test-jar>false</no-test-jar>
         <workflow-step-api.version>2.13</workflow-step-api.version>
--- a/pom.xml
+++ b/pom.xml
@@ -67,12 +67,13 @@ THE SOFTWARE.
         <changelist>-SNAPSHOT</changelist>
         <java.level>8</java.level>
         <jenkins.version>2.138.4</jenkins.version>
-        <scm-api.version>2.2.7</scm-api.version>
+        <scm-api.version>2.6.3</scm-api.version>
         <no-test-jar>false</no-test-jar>
         <workflow-step-api.version>2.13</workflow-step-api.version>
         <workflow-support.version>2.17</workflow-support.version>
         <workflow-cps.version>2.53</workflow-cps.version>
         <git-plugin.version>3.9.1</git-plugin.version>
+        <subversion-plugin.version>2.13.0</subversion-plugin.version>
     </properties>
     <dependencies>
         <dependency>
@@ -201,7 +202,7 @@ THE SOFTWARE.
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>subversion</artifactId>
-            <version>2.7.1</version>
+            <version>${subversion-plugin.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
@@ -224,7 +225,7 @@ THE SOFTWARE.
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>subversion</artifactId>
-            <version>2.7.1</version>
+            <version>${subversion-plugin.version}</version>
             <classifier>tests</classifier>
             <scope>test</scope>
         </dependency>
@@ -256,7 +257,7 @@ THE SOFTWARE.
         <dependency>
             <groupId>org.jenkins-ci.modules</groupId>
             <artifactId>instance-identity</artifactId>
-            <version>2.1</version>
+            <version>2.2</version>
             <scope>test</scope>
         </dependency>
         <dependency>
@@ -268,7 +269,7 @@ THE SOFTWARE.
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>credentials</artifactId>
-            <version>2.1.14</version>
+            <version>2.1.16</version>
             <scope>test</scope>
         </dependency>
         <dependency>


slonopotamus:
@jglick Hmm... `SCMBinderTest.exactRevisionSubversion` [doesn't pass on Linux](https://ci.jenkins.io/job/Plugins/job/workflow-multibranch-plugin/job/PR-91/2/testReport/junit/org.jenkinsci.plugins.workflow.multibranch/SCMBinderTest/exactRevisionSubversion/).

pom.xml
```diff
@@ -66,8 +66,7 @@ THE SOFTWARE.
         <revision>2.22</revision>
         <changelist>-SNAPSHOT</changelist>
         <java.level>8</java.level>
-        <jenkins.version>2.121.1</jenkins.version>
```
jglick: You should always specify an explicit `jenkins.version`. Do _not_ attempt to inherit from the parent POM; that is merely a minimum recommended version.

jglick:
Seems that `SCMBinderTest.exactRevisionSubversion` passes with a `subversion` dep up to `2.7.2`. From `2.8`–`2.11.1` it fails with

```
org.junit.ComparisonFailure: expected:<[[/prj/trunk/Jenkinsfile, /prj/trunk/]file]> but was:<[[Jenkinsfile, ]file]>
	at org.junit.Assert.assertEquals(Assert.java:115)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at org.jenkinsci.plugins.workflow.multibranch.SCMBinderTest.exactRevisionSubversion(SCMBinderTest.java:167)
```

and from `2.12.0`–`2.13.0` it fails with

```
java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
	at java.util.ArrayList.rangeCheck(ArrayList.java:657)
	at java.util.ArrayList.get(ArrayList.java:433)
	at org.jenkinsci.plugins.workflow.multibranch.SCMBinderTest.exactRevisionSubversion(SCMBinderTest.java:160)
```

I will dig into this a bit and probably propose a move of the SVN-based tests out of here. (They do exercise some aspects of Pipeline/SCM integration that the Git plugin does not, due to the very different model of versioning, but probably not enough to be worth the hassle of testing against what is increasingly an obsolete tool.)

pom.xml
slonopotamus: Ok, reverted this change.

pom.xml
```diff
@@ -67,7 +67,7 @@ THE SOFTWARE.
         <changelist>-SNAPSHOT</changelist>
         <java.level>8</java.level>
         <jenkins.version>2.121.1</jenkins.version>
```
jglick: May work but 2.138.4 is the minimum recommended version.

pom.xml
```diff
@@ -201,7 +201,7 @@ THE SOFTWARE.
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>subversion</artifactId>
-            <version>2.7.1</version>
+            <version>2.13.0</version>
```
jglick: Need to introduce a POM property so that the `tests` JAR version (line ~226) is also updated.

slonopotamus: Good catch, I missed that.

pom.xml
slonopotamus: Okay.

pom.xml
slonopotamus: Done

jglick:
BTW please do not force-push inside a PR branch! It breaks the GH review flow, makes regular `git` CLI operations like `merge` and `pull` break in crazy ways, etc. The repository maintainer may choose to squash-merge or whatever if and when the PR is accepted, but leave that decision to them.

jglick:
also need

```diff
         <dependency>
             <groupId>org.tmatesoft.svnkit</groupId>
             <artifactId>svnkit-cli</artifactId>
-            <version>1.8.14</version>
+            <version>1.10.1</version>
             <scope>test</scope>
         </dependency>
         <dependency>
```

jglick:
The earlier test failure is probably from https://github.com/jenkinsci/subversion-plugin/pull/183 (unconfirmed), which seems harmless enough, but the later one looks like a real regression so far. I will continue working to track it down as de facto maintainer (not that I want to be) but I think this PR is hopeless to continue as such; I will likely just wind up filing a fresh PR to handle the test split once the root causes are understood.

jglick:
OK that might be an overreaction. The second failure looks to be from https://github.com/jenkinsci/subversion-plugin/pull/223 and is unavoidable pending more complex changes to `scm-api` to allow lightweight changelog calculation. So for now it suffices, in addition to the `svnkit-cli` bump above (needed to match upstream), to do

```diff
diff --git src/test/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinderTest.java src/test/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinderTest.java
index 3e418bd..401c37a 100644
--- src/test/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinderTest.java
+++ src/test/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinderTest.java
@@ -48,6 +48,7 @@ import jenkins.scm.api.SCMHead;
 import jenkins.scm.api.SCMRevision;
 import jenkins.scm.api.SCMRevisionAction;
 import jenkins.scm.api.SCMSourceDescriptor;
+import jenkins.scm.impl.subversion.SubversionSCMFileSystem;
 import jenkins.scm.impl.subversion.SubversionSCMSource;
 import static org.hamcrest.Matchers.*;
 
@@ -129,6 +130,9 @@ public class SCMBinderTest {
         }
     }
 
+    static {
+        System.setProperty(SubversionSCMFileSystem.DISABLE_PROPERTY, "true");
+    }
     @Test public void exactRevisionSubversion() throws Exception {
         sampleSvnRepo.init();
         ScriptApproval sa = ScriptApproval.get();
@@ -164,7 +168,7 @@ public class SCMBinderTest {
         assertTrue(iterator.hasNext());
         ChangeLogSet.Entry entry = iterator.next();
         assertEquals("tweaked", entry.getMsg());
-        assertEquals("[/prj/trunk/Jenkinsfile, /prj/trunk/file]", new TreeSet<>(entry.getAffectedPaths()).toString());
+        assertEquals("[Jenkinsfile, file]", new TreeSet<>(entry.getAffectedPaths()).toString());
         assertFalse(iterator.hasNext());
     }
 
```

With that I think this could be mergeable, and then an analogue to https://github.com/jenkinsci/workflow-multibranch-plugin/pull/17 could be done as a follow-up.

jglick:
BTW thanks for your patience on this! The real problem is that the Jenkins project does not consistently run https://github.com/jenkinsci/plugin-compat-tester which would allow us to detect integration test failures and genuine regressions in a timely manner, so people doing routine housekeeping wind up pulling in years’ worth of technical debt.

slonopotamus:
I'm afraid I'm not as patient as you think.

You have #90 that passes tests, but is *a bit* hackish. You have this PR that is cleaner but exposes the fact that workflow-multibranch-plugin doesn't play together with modern subversion-plugin.

The only problem I wanted to fix - bugs that are related to updating to new parent pom and, specifically, new JTH with jenkinsci/jenkins-test-harness#166 that affected several plugins. workflow-multibranch-plugin is one of the plugins that has issues. I spent time reproducing the problem, I found the root cause, I gave a solution.

But dealing with consequences of global Jenkins code organization decisions that lead to plugins running tests against randomly-outdated versions of each other is beyound of what I am ready to handle now.

jglick:
Ack. I can just refile this plus test fixes.

PR #92, opened 2019-12-18
Updates

jglick:
Continuation of #91 by @slonopotamus, picking up https://github.com/jenkinsci/subversion-plugin/pull/240 while adapting to https://github.com/jenkinsci/subversion-plugin/pull/183 + https://github.com/jenkinsci/subversion-plugin/pull/223, all originally (?) to adapt to https://github.com/jenkinsci/jenkins-test-harness/pull/166. Follow-ups would be

* extract SVN tests to `subversion-plugin`, analogous to #17
* switch most dependencies to the BOM

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinderTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinderTest.java
@@ -48,6 +48,7 @@ import jenkins.scm.api.SCMHead;
 import jenkins.scm.api.SCMRevision;
 import jenkins.scm.api.SCMRevisionAction;
 import jenkins.scm.api.SCMSourceDescriptor;
+import jenkins.scm.impl.subversion.SubversionSCMFileSystem;
 import jenkins.scm.impl.subversion.SubversionSCMSource;
 import static org.hamcrest.Matchers.*;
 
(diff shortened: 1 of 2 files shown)


slonopotamus:
To be more exact, `workflow-multibranch-plugin` only needs [newer `mapdb-api`](https://github.com/jenkinsci/subversion-plugin/pull/240/commits/6b9896eba70c0ac4b52c78fcfade692b43756f04) in order to adapt for jenkinsci/jenkins-test-harness#166. Either by doing it itself (as was done in #90) or by bumping `subversion` to >= 2.13.0. Latter makes more sense because bogus `mapdb-api` is a dependency of `subversion`, `workflow-multibranch-plugin` doesn't want to know anything about `mapdb`.

src/test/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinderTest.java
```diff
@@ -129,6 +130,9 @@ public static void assertRevisionAction(WorkflowRun build) {
         }
     }
 
+    static {
+        System.setProperty(SubversionSCMFileSystem.DISABLE_PROPERTY, "true");
+    }
```
dwnusbaum: Do we need to tear this down in between tests, or do we always want it to be `true`, so it doesn't really matter?

jglick: Should not really matter I think.

jglick:
BTW locally I am getting a failure from `ReplayActionTest.scriptFromSCM` though it seems to work on CI:

    A branch named 'master' already exists.

Not going to worry about it right now, and unrelated to `subversion` (this is a `git`-based test).

PR #94, opened 2020-03-01
[JENKINS-45273] Allow usage of Jenkinsfile defined elsewhere

fcristovao:
Hey there,

It's my first PR towards Jenkins CI related stuff, so please let me know if I missed anything in terms of process.

This PR fixes [JENKINS-45273](https://issues.jenkins-ci.org/browse/JENKINS-45273). I've seen people asking for this feature on [Stack Overflow](https://stackoverflow.com/questions/39041342/jenkins-multibranch-pipeline-with-jenkinsfile-from-different-repository), and there's [even another plugin](https://github.com/jenkinsci/remote-file-plugin) that tries to do exactly this.
However, in [the documentation](https://wiki.jenkins.io/display/JENKINS/Before+starting+a+new+plugin), it suggests to add contribute to the original plugin instead, so I thought this would be a better alternative.


dwnusbaum:
@fcristovao Thanks for the PR! Given that https://github.com/jenkinsci/remote-file-plugin (which unfortunately has a very non descriptive name) already exists, I think it would make more sense for any related work to be done in that plugin going forward, so that we don't have two plugins with the same functionality. I think the documentation is mostly trying to avoid situations where someone totally reimplements a plugin or feature from scratch, but in this case, Remote File plugin uses this plugin as a dependency and adds an extension with distinct functionality from the base plugin, so it seems ok.

I will  go ahead close JENKINS-45273 with a reference to the Remote File Plugin.

fcristovao:
I thought that the point was not to have an explosion of Plugins, so I thought it would be better to incorporate these changes in the original plugin instead?
My worry is that otherwise the plugins would diverge and eventually be incompatible.
Also, it seems that that plugin doesn't "marry" very well with other extension points?

dwnusbaum:
> I thought that the point was not to have an explosion of Plugins, so I thought it would be better to incorporate these changes in the original plugin instead

I think there are two cases here with distinct guidance:

1. You want to reimplement the main functionality of an existing plugin. In this case, it is ideal to modify the existing plugin rather than create a new plugin that does the same thing. For example, we don't want "Pipeline: Multibranch Plugin" and "Pipeline: Cooler Multibranch Plugin" to exist simultaneously with the same feature set, forcing users to choose between one or the other.
2. You want to add a new feature that is related to an existing plugin (for example, to implement an `ExtensionPoint` defined in the plugin). In this case, the preferred approach can vary, but in general, it is totally reasonable and often ideal to implement the extension in a new plugin, so that users can install only the extensions that are actually useful to them. Adding a bunch of often-unused features to the main plugin can make it much harder to maintain, and it can make things more confusing for users by showing them a bunch of options that they may not care about.

For an example of case 2, shared libraries (from Pipeline: Shared Groovy Libraries Plugin) only implement retrieval from SCMs by default, because this was seen as the most common use case. Other plugins can implement a custom extension for the `LibraryRetriever` `ExtensionPoint`, such as [Pipeline: Shared Groovy Libraries through HTTP retrieval Plugin](https://plugins.jenkins.io/workflow-cps-global-lib-http/), which allows a library to be retrieved directly from an HTTP endpoint.

> My worry is that otherwise the plugins would diverge and eventually be incompatible

Yes, the maintainers of any plugin that depends on another have to make sure that their plugin is compatible with the upstream plugin, but I think this is reasonable. This is usually not too difficult, because maintainers of plugins with extension points that are known to be in use by other plugins try to avoid making any gratuitous changes that would break compatibility with the downstream plugins.

> Also, it seems that that plugin doesn't "marry" very well with other extension points?

Yes, this is unfortunate, if we had known about Remote File Plugin before it was made, we could have tried to guide it to use a different approach that worked better with the existing extension points. Even so, I would first try to modify that plugin to make it work better rather than recreating it. (Also, I would encourage the developers to change the plugin's display name to something like "Multibranch Pipeline Remote Jenkinsfile Plugin" for clarity)

fcristovao:
Ok then, I'll try to merge in that Plugin Repo instead.

PR #95, opened 2020-04-13
JENKINS-61489 Adding french locale

jbleduigou:
Implementing https://issues.jenkins-ci.org/browse/JENKINS-61489

Signed-off-by: Jean-Baptiste Le Duigou <jb.leduigou@gmail.com>

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/multibranch/Messages_fr.properties
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/multibranch/Messages_fr.properties
@@ -1,0 +1,3 @@
+    ReadTrustedStep._has_been_modified_in_an_untrusted_revis=\u2018{0}\u2019 a été modifié dans une révision non fiable
+WorkflowMultiBranchProject.DisplayName=Pipeline Multibranches
+WorkflowMultiBranchProject.Description=Cr00e9e un ensemble de projets Pipeline en se basant sur les branches d00e9tect00e9es dans le d00e9p00f4t d'un gestionnaire de code source.
--- a/src/main/resources/org/jenkinsci/plugins/workflow/multibranch/Messages_fr.properties
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/multibranch/Messages_fr.properties
@@ -1,3 +1,3 @@
-    ReadTrustedStep._has_been_modified_in_an_untrusted_revis=\u2018{0}\u2019 a été modifié dans une révision non fiable
+ReadTrustedStep._has_been_modified_in_an_untrusted_revis=\u2018{0}\u2019 a \u00e9t\u00e9 modifi\u00e9 dans une r\u00e9vision non fiable
 WorkflowMultiBranchProject.DisplayName=Pipeline Multibranches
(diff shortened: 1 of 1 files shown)


src/main/resources/org/jenkinsci/plugins/workflow/multibranch/Messages_fr.properties
```diff
@@ -0,0 +1,3 @@
+    ReadTrustedStep._has_been_modified_in_an_untrusted_revis=\u2018{0}\u2019 a été modifié dans une révision non fiable
+WorkflowMultiBranchProject.DisplayName=Pipeline Multibranches
+WorkflowMultiBranchProject.Description=Cr00e9e un ensemble de projets Pipeline en se basant sur les branches d00e9tect00e9es dans le d00e9p00f4t d'un gestionnaire de code source.
```
dwnusbaum, line 3: Some of the digit sequences here look like they were intended to be escape sequences or something, for example, is `00e9` supposed to be `é`?

jbleduigou: There are indeed some problems here.
I will fix them and update the PR.

dwnusbaum:
@jbleduigou Thanks for updating the PR! Just so you know, if you plan on adding French localizations for a lot of plugins, the best approach is to create a plugin for the French locale, and then all of the translations can live in one place, and the maintainers of that plugin can all be fluent French speakers to facilitate PR review. Here is an example of a plugin like that https://github.com/jenkinsci/localization-zh-cn-plugin in case that sounds like something you might want to create.

src/main/resources/org/jenkinsci/plugins/workflow/multibranch/Messages_fr.properties
```diff
@@ -0,0 +1,3 @@
+ReadTrustedStep._has_been_modified_in_an_untrusted_revis=\u2018{0}\u2019 a \u00e9t\u00e9 modifi\u00e9 dans une r\u00e9vision non fiable
+WorkflowMultiBranchProject.DisplayName=Pipeline Multibranches
+WorkflowMultiBranchProject.Description=Cr\u00e9e un ensemble de projets Pipeline en se basant sur les branches d\u00e9tect\u00e9es dans le d\u00e9p\u00f4t d'un gestionnaire de code source.
```
batmat, line 3: Side note: give `.properties` are by definition encoded in `ISO-8859-1`, IMO it's probably unnecessary to encode using unicode entities. It is harder to review and hence easier to miss typos and the likes.

Thanks!

jbleduigou:
> @jbleduigou Thanks for updating the PR! Just so you know, if you plan on adding French localizations for a lot of plugins, the best approach is to create a plugin for the French locale, and then all of the translations can live in one place, and the maintainers of that plugin can all be fluent French speakers to facilitate PR review. Here is an example of a plugin like that https://github.com/jenkinsci/localization-zh-cn-plugin in case that sounds like something you might want to create.

Thanks for reviewing and for your comment!
That is definitely an approach that seems better.
I would look into it in the next days.

PR #100, opened 2020-07-31
Update minimum core version, parent POM, and dependencies

dwnusbaum:
Had this on the back burner waiting for https://github.com/jenkinsci/bom/pull/264 and https://github.com/jenkinsci/bom/pull/278 to make it into a release, but going ahead and filing it now so we can release `workflow-multibranch` to unblock https://github.com/jenkinsci/bom/pull/280.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMVar.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMVar.java
@@ -97,7 +97,7 @@ import org.jenkinsci.plugins.workflow.support.pickles.XStreamPickle;
         } else {
             SCMHead head = branch.getHead();
             FlowExecutionOwner owner = ((WorkflowRun) build).asFlowExecutionOwner();
-            TaskListener listener = owner != null ? owner.getListener() : TaskListener.NULL;
+            TaskListener listener = owner.getListener();
             tip = scmSource.fetch(head, listener);
             if (tip == null) {
                 throw new AbortException("Could not determine exact tip revision of " + branch.getName());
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMVar.java
```diff
@@ -97,7 +97,7 @@
         } else {
             SCMHead head = branch.getHead();
             FlowExecutionOwner owner = ((WorkflowRun) build).asFlowExecutionOwner();
-            TaskListener listener = owner != null ? owner.getListener() : TaskListener.NULL;
+            TaskListener listener = owner.getListener();
```
dwnusbaum, line 100: `asFlowExecutionOwner` is `@NonNull` in recent versions of `workflow-job`, so SpotBugs didn't like the redundant null check.

src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
```diff
@@ -254,8 +254,10 @@ public void testChoiceParameterSnippetizer() throws Exception {
         r.waitUntilNoActivity(); // #1 built automatically
         assertEquals(1, p.getBuilds().size());
         r.assertBuildStatusSuccess(p.scheduleBuild2(0)); // #2
+        Thread.sleep(500L); // WorkflowRun performs log rotation asynchronously.
```
dwnusbaum, line 257: This test was failing consistently on my machine after the updates. I did not see a trivial way to wait for [this task](https://github.com/jenkinsci/workflow-job-plugin/blob/7a9d13220aacd4b535cc70976f8af082e7820644/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java#L618-L624) to complete before continuing and so just added a  call to `Thread.sleep`...

dwnusbaum:
There seems to be something wrong with the normal Linux agents on ci.jenkins.io (the build has been waiting on an agent for hours), so I pushed a commit to use ACI agents to see if that helps.

PR #128, opened 2021-06-12
Loosening `JobPropertyStepTest#configRoundTripBuildDiscarder`.

basil:
See jenkinsci/bom#471 and jenkinsci/bom#472. These PRs are attempting to add new plugins to the BOM, which trip up `JobPropertyStepTest#configRoundTripBuildDiscarder`. `JobPropertyStepTest#configRoundTripBuildDiscarder` expects to see exactly one job property after round-tripping the configuration. But when those other two plugins are on the classpath, round-tripping has three job properties:

```
0 = {BuildDiscarderProperty@11636}
1 = {RebuildSettings@11637}
2 = {ThrottleJobProperty@11638}
```

This PR simply loosens the test to allow for additional job properties to be present. This allows PCT to pass when Rebuild and Throttle Concurrent Builds are also on the classpath.

PR #152, opened 2022-01-05
Extend PCT fix to `JobPropertyStepTest.configRoundTripParameters`

jglick:
#128 patched one test for `bom` PCT but neglected to do the same for a similar test. Fixes https://github.com/jenkinsci/bom/issues/811.


timja (approved):

basil (approved):
Thank you!

basil (approved):

PR #159, opened 2022-02-16
Security fix test compatibility

dwnusbaum:
The tests for the security fixes for https://www.jenkins.io/security/advisory/2022-02-15/ fail on Windows due to differences in behavior. Two currently ignored tests (they need an updated workflow-cps dependency) would also fail against recent versions of Git plugin due to https://github.com/jenkinsci/git-plugin/pull/1207, so I fixed that at the same time.

- [x] Link to relevant pull requests, esp. upstream and downstream changes

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/multibranch/ReadTrustedStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/multibranch/ReadTrustedStepTest.java
@@ -322,7 +322,10 @@ public class ReadTrustedStepTest {
         FileUtils.copyDirectory(new File(sampleRepo.getRoot(), ".git"), gitDirInSvnRepo);
         String jenkinsRootDir = r.jenkins.getRootDir().toString();
         // Add a Git post-checkout hook to the .git folder in the SVN repo.
-        Files.write(gitDirInSvnRepo.toPath().resolve("hooks/post-checkout"), ("#!/bin/sh\ntouch '" + jenkinsRootDir + "/hook-executed'\n").getBytes(StandardCharsets.UTF_8));
+        Path postCheckoutHook = gitDirInSvnRepo.toPath().resolve("hooks/post-checkout");
+        // Always create hooks directory for compatibility with https://github.com/jenkinsci/git-plugin/pull/1207.
+        Files.createDirectories(postCheckoutHook.getParent());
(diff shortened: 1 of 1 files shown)


src/test/java/org/jenkinsci/plugins/workflow/multibranch/ReadTrustedStepTest.java
```diff
@@ -232,12 +232,13 @@ public void pathTraversalRejected() throws Exception {
 
         WorkflowRun b = p.getLastBuild();
         assertEquals(1, b.getNumber());
-        r.assertLogContains("secrets/master.key references a file that is not inside " + r.jenkins.getWorkspaceFor(p).getRemote(), b);
+        r.assertLogContains("master.key references a file that is not inside " + r.jenkins.getWorkspaceFor(p).getRemote(), b);
```
dwnusbaum, line 235: I think this was just an issue with `secrets\master.key` vs `secrets/master.key`

src/test/java/org/jenkinsci/plugins/workflow/multibranch/ReadTrustedStepTest.java
```diff
@@ -259,6 +260,7 @@ public void symlinksInReadTrustedCannotEscapeWorkspaceContext() throws Exception
     @Issue("SECURITY-2491")
     @Test
     public void symlinksInUntrustedRevisionCannotEscapeWorkspace() throws Exception {
+        assumeFalse(Functions.isWindows()); // On Windows, the symlink is treated as a regular file, so there is no vulnerability, but the behavior is different.
```
dwnusbaum, line 263: In other words, the contents of the symlink itself get read, not the contents of the file that it links to.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/ReadTrustedStep.java
```diff
@@ -248,7 +248,7 @@ private static boolean isDescendant(FilePath child, FilePath parent) throws IOEx
             if (child.isRemote() || parent.isRemote()) {
                 throw new IllegalStateException();
             }
-            return new File(child.getRemote()).getCanonicalFile().toPath().startsWith(parent.absolutize().getRemote());
+            return new File(child.getRemote()).getCanonicalFile().toPath().startsWith(new File(parent.getRemote()).getCanonicalPath());
```
dwnusbaum, line 251: The only intended difference here is that if `JENKINS_HOME` is a symlink, the code should no longer throw an exception in non-dangerous scenarios.

jglick (approved):

PR #219, opened 2022-12-14
Bump baseline, parent, BOM, etc.

jglick:
General metadata refresh. Supersedes #196, #198, #211, #213, & #217.


raul-arabaolaza (approved):

Jenkinsfile
```diff
@@ -1,5 +1,6 @@
-buildPlugin(useContainerAgent: true, configurations: [
-    [ platform: "linux", jdk: "8" ],
-    [ platform: "windows", jdk: "8" ],
-    [ platform: "linux", jdk: "11" ]
+buildPlugin(
+  useContainerAgent: true,
+  configurations: [
+    [platform: 'linux', jdk: 17],
+    [platform: 'windows', jdk: 11],
```
jtnord, line 5: personally I do not like this idiom.  Should change one variable so if something breaks you have a better idea of the cause.
is it java11 or is it windows.
the windows agents are generally less available so you still have to wait around anyway (or they are just broken in which case you have no idea if it even works on java 11)

```suggestion
    [platform: 'linux', jdk: 17],
    [platform: 'linux', jdk: 11],
    [platform: 'windows', jdk: 11],
```

jtnord (approved):

PR #220, opened 2022-12-14
[JENKINS-46795] Abort builds with untrusted `Jenkinsfile`, but only given passive cause

jglick:
[JENKINS-46795](https://issues.jenkins.io/browse/JENKINS-46795) amending #69. Subsumes #219. Note that some much less common Jenkins features use `getTrustedRevision` and would not pick up this logic: https://github.com/jenkinsci/pipeline-groovy-lib-plugin/blob/b44ce045b582f60e5b76d4a9c4cbf52dfa4184de/src/main/java/org/jenkinsci/plugins/workflow/libs/MultibranchScmRevisionVerifier.java#L62 and https://github.com/jenkinsci/basic-branch-build-strategies-plugin/blob/a76ef8f7fcb9ee0c97b9f8408c1c11e0acd29154/src/main/java/jenkins/branch/buildstrategies/basic/ChangeRequestBuildStrategyImpl.java#L138


jtnord (commented):
as per the jira
> It should override checkTrusted(GitHubSCMSourceRequest, PullRequestSCMRevision) to check the GitHub API to see if the current revision has been approved by a maintainer.


IIUC approval is at the PR level not the commit level.
so making this change would introduce a security issue for all existing setups where the user has not setup the github branch to dismiss approvals on push?

If that is the case then this needs to be opt in and not the default behaviour.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java
```diff
@@ -143,6 +153,24 @@ public SCMBinder(String scriptPath) {
         return new CpsScmFlowDefinition(scm, scriptPath).create(handle, listener, actions);
     }
 
+    private static Set<Class<? extends Cause>> passiveCauses = Set.of(
+        BranchIndexingCause.class,
+        BranchEventCause.class,
+        SCMTrigger.SCMTriggerCause.class,
+        TimerTrigger.TimerTriggerCause.class);
```
jtnord, line 160: given Jenkins extensibility, would it not be better to have a list of non passive causes to prevent any security issues in future?

jtnord: e.g. upstream / downstream cause?  Jira comment?
if using `withMaven` what if a new SNAPSHOT dependency was built elsewhere

jglick: > a list of non passive (allowed) causes

Considered that, though it seemed more intrusive since for example `github-checks` (and the CloudBees CI equivalent) would both have to add to that list, which means either hard-coding class names or introducing a new extension point (in `scm-api`?) and having those plugins implement it. Something to consider.

> upstream / downstream cause?

Should not be triggering PR builds; or if it were, presumably for a reason (some other project called `build` on this PR branch project).

> Jira comment?

Not sure what you refer to.

> if using `withMaven` what if a new `SNAPSHOT` of a dependency was built elsewhere

Hmm, will need to check if that is in fact possible.

jtnord: > > Jira comment?
> Not sure what you refer to.

some jira plugin that allows to trigger builds based off a comment in Jira.

jglick:
> this code does not "trust PR revisions when approved" which is what the Jira title is. That is confusing and should be changed

Right, the original proposal in the Jira description would have required changes to every branch source with a notion of an untrusted change request, and furthermore would have required expanded webhooks since Jenkins would need to listen to the PR being approved after it had already rejected the attempt to build the commit the first time, _and_ would not anyway offer repository owners as much flexibility since they might wish to see CI results without necessarily approving the PR for final merge. (OTOH it would allow triggering CI without App authn or a Jenkins login.)

jglick:
> not so sure about not trusting causes based on a denylist rather than an include list

Putting into draft while I investigate options.

jglick:
Tested interactively:
* ran this PR hooked up to an App with webhook via ngrok
* set up a multibranch project (PR-head strategy) with default trust settings
* filed a PR as a user with no write permission editing `Jenkinsfile` ⇒ check failed as expected
* used **Build Now** from Jenkins web UI ⇒ passed (using edited `Jenkinsfile`)
* used **Replay** after that ⇒ passed (note that you cannot currently replay an aborted build because it never got to the point of loading a `Jenkinsfile`)
* edited PR as author ⇒ failed as expected
* used **Re-run check** from GH UI ⇒ passed (using edited `Jenkinsfile`)

src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java
```diff
@@ -165,22 +167,4 @@ public SCMBinder(String scriptPath) {
 
     }
 
-    // TODO seems there is no general-purpose ConsoleNote which simply wraps markup in specified HTML
-    @SuppressWarnings("rawtypes")
-    public static class WarningNote extends ConsoleNote {
```
daniel-beck, line 170: `@Restricted(NoExternalUse.class)`?

daniel-beck (commented):
Since this relaxes security to support the use case of `Jenkinsfile` contributions from untrusted contributors, I strongly recommend this be accompanied by end-user documentation. Ideally there'd be a place in the UI for that, but I'm not sure where that would be, except perhaps wordier log messages when _not_ trusting, to point out that manual re-builds would result in trust.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java
jglick: I do not see why; it was not used elsewhere to begin with.

jglick:
> wordier log messages when _not_ trusting

https://github.com/jenkinsci/scm-api-plugin/pull/180/commits/f6ec0d005bb5d6ee1288f54cbd105dad86af42d2 should help some. I agree that this sort of thing deserves more formal documentation, though. Can think about suitable spots in the GUI, and/or jenkins.io/doc.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java
daniel-beck: That said, removing this class might cause trouble with old build logs containing non-deserializable annotations? Or is this going somewhere else?

jglick: https://github.com/jenkinsci/jenkins/blob/aa83562568b44aa85ff17bdfab831bbe39c2fd15/core/src/main/java/hudson/console/ConsoleAnnotationOutputStream.java#L112-L125 should just drop the annotation, i.e. make the text be plain rather than yellow. Not specifically tested.

daniel-beck:
Given e.g. occasional infra hiccups necessitating rebuilds of projects for reasons outside the actual code and build script, would it make sense (or even be feasible) to limit elevating trust to situations when the previous build failed because of this lack of trust? Or would that overcomplicate things?

Thinking of cases when an admin notices an infra problem (potentially even their fault) and then checks recent builds, triggering rebuilds of any that had failed, to least inconvenience the developers. That would be something we'd need to discourage with the current implementation.

Perhaps we could make this at least explicit, e.g. with `AlternativeUiTextProvider` labeling the button as "Trust and Build Now" or so?

jglick:
> with `AlternativeUiTextProvider` labeling the button as "Trust and Build Now"

:thinking: could be possible with a custom `CauseOfInterruption` if `workflow-cps` were made to look for that and offer a different version of **Replay**. In that case I suppose rather than trusting `UserIdCause` we would trust the new `Cause` added by that gesture (in addition to `UserIdCause`).

Integration of `github-checks` would remain as is: **Re-run** from **Checks** would be considered trusted regardless of original cause of failure.

daniel-beck:
Here's an idea for re-building from the Jenkins UI: How about an `Action` attached to the aborted build, that rebuilds it with trust (e.g. a new cause, perhaps a `UserIdCause` subclass if it's non-`final`?), and it has separate UI? That way, you can look at the log and click a link or button right there in the "I'm rejecting this" message (not unlike the "Terminate all running steps" etc. messages), or have a new sidepanel link for the build (in addition to Build Now to minimize "rebuild all" risk), or a button on the main build page? This way, no need to place trust in the basic `UserIdCause`, and admins trying to be helpful are not a security risk.

> Integration of `github-checks` would remain as is: **Re-run** from **Checks** would be considered trusted regardless of original cause of failure.

Does GitHub support multiple different "Re-run" gestures, with different behavior, and to what extent can the rejection of the `Jenkinsfile` be reported clearly to GH checks? Maybe this is clear (haven't gotten around to manually running this yet), so if the result reported to GH is unambiguous (e.g., marked as aborted, with log excerpt, etc.), this should be fine. The only potential issue I see would be in case of unclear results reported to the GH check ("Failed" status same as any failed build or infra hiccup, no or unclear log excerpt etc.), as it may be difficult to tell that a complex PR changing lots of files also changes the `Jenkinsfile`.

jglick:
> look at the log and click a link or button right there in the "I'm rejecting this" message

Yes, that would be helpful for users of the classic log display.

> a new sidepanel link for the build

Right, basically the

> different version of **Replay**

that I implied in https://github.com/jenkinsci/workflow-multibranch-plugin/pull/220#issuecomment-1508819962.

> Does GitHub support multiple different "Re-run" gestures, with different behavior

No. You can attach “actions” to a check but these have a completely different UX. IIRC you cannot even control whether **Re-run** is displayed; if a check fails, the link is shown, and if clicked, sends an event which the App can respond to or not.

> to what extent can the rejection of the `Jenkinsfile` be reported clearly to GH checks?

It has to be displayed as a “failure” (unfortunately the “cancelled” conclusion does not offer **Re-run**), but the associated text can say whatever you want. Would require some hacking around in `github-checks` to specialize the display in this way, however, and currently it is not straightforward to make changes to that plugin because of its nonstandard POM (https://github.com/jenkinsci/github-checks-plugin/pull/307).

daniel-beck:
> > look at the log and click a link or button right there in the "I'm rejecting this" message
> 
> Yes, that would be helpful for users of the classic log display.
> 
> > a new sidepanel link for the build
> 
> Right, basically the
> 
> > different version of **Replay**
> 
> that I implied in [#220 (comment)](https://github.com/jenkinsci/workflow-multibranch-plugin/pull/220#issuecomment-1508819962).

👍 

> > to what extent can the rejection of the `Jenkinsfile` be reported clearly to GH checks?
> 
> It has to be displayed as a “failure” (unfortunately the “cancelled” conclusion does not offer **Re-run**), but the associated text can say whatever you want. Would require some hacking around in `github-checks` to specialize the display in this way, however, and currently it is not straightforward to make changes to that plugin because of its nonstandard POM ([jenkinsci/github-checks-plugin#307](https://github.com/jenkinsci/github-checks-plugin/pull/307)).

Would it be a viable step forward to proceed with support for Jenkins UI only for now, if the state of `github-checks` is a problem in making the integration with this failure nicer?

jglick:
> Would it be a viable step forward to proceed with support for Jenkins UI only

I do not think so. At least, it would make this feature useless for ci.jenkins.io, since most repository maintainers lack **Job/Build** permissions on the server.

> if the state of `github-checks` is a problem in making the integration with this failure nicer

IMO just making **Re-run** be considered trusted, as is already done upstream, suffices. This would be unrelated to the concern you raised in https://github.com/jenkinsci/workflow-multibranch-plugin/pull/220#issuecomment-1508148233.

daniel-beck:
> IMO just making **Re-run** be considered trusted, as is already done upstream, suffices. This would be unrelated to the concern you raised in https://github.com/jenkinsci/workflow-multibranch-plugin/pull/220#issuecomment-1508148233.

Right. Re-run is less prone to admin accidents than Build Now, although complex PRs might still manage to trick reviewers into assuming flakiness when it's actually lack of trust.

I do not object to going ahead with trust in Re-Run, though IMO it would be very useful to be able to use this functionality from within Jenkins as well.

PR #221, opened 2022-12-15
add java 11 back for linux

jtnord:
I missed that @jglick had enabled auto merge in #219 so the [feedback](https://github.com/jenkinsci/workflow-multibranch-plugin/pull/219#discussion_r1049496731) did not even get looked at before it was merged.

ideally you should only change one variable at once so if something breaks you have a better idea of the cause. is it java11 or is it windows.
the windows agents are generally less available so you still have to wait around anyway (or they are just broken in which case you have no idea if it even works on java 11)

- [x] Link to relevant pull requests, esp. upstream and downstream changes

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,5 +2,6 @@ buildPlugin(
   useContainerAgent: true,
   configurations: [
     [platform: 'linux', jdk: 17],
+    [platform: 'linux', jdk: 11],
     [platform: 'windows', jdk: 11],
 ])


jglick (dismissed):
Deliberately skipping the redundant matrix combination. This would add 50% to CI load, and in practice it is very unlikely that a problem would appear on Linux but not Windows _and_ appear on Java 11 but not Java 17. https://github.com/jenkinsci/archetypes/pull/551

jglick:
Put another way:

Windows CI builds fail pretty frequently. When the reason is specific to Windows, or at least the CI environment we have available, it is usually obvious enough.

In my experience using Java 17 is unlikely to introduce a build failure, provided you are using reasonably recent tooling. If there is some illegal use of reflection, that will be clear. Other than that, things are generally very compatible.

jtnord:
> Windows CI builds fail pretty frequentl

yes but more often because the channel dies breaks or the agent dies rather than a test failure in my experience.  Many PRs are getting merged with "windows agent failure" leading to zero java11 coverage which is a big mistake

basil:
https://github.com/jenkinsci/archetypes/pull/547#discussion_r1041327868

jglick:
> more often because the channel dies breaks or the agent dies rather than a test failure

Even after https://github.com/jenkins-infra/pipeline-library/pull/405?

> PRs are getting merged with "windows agent failure"

The check should be rerun until all are green before merging.

jglick:
In sum, I continue to believe that the Linux/17 + Windows/11 matrix combination is a good tradeoff between test coverage, transparency, and efficiency of ci.jenkins.io resources. If multiple people feel strongly otherwise (and ci.jenkins.io admins do not object to the added load), propose to change the recommended defaults in `archetypes` and the matching copy in jenkins.io/doc/developer.

PR #229, opened 2023-02-07
Adding symbol to the WorkflowMultiBranchProject class descriptor

joseblas:
The WorkflowMultiBranchProject class does not have a proper Symbol in the descriptor as many other such as FreeStyleProject (https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/model/FreeStyleProject.java#L79)

- [ ] Ensure that the pull request title represents the desired changelog entry

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProject.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProject.java
@@ -134,7 +134,7 @@ public class WorkflowMultiBranchProject extends MultiBranchProject<WorkflowJob,W
                 strategy != null ? strategy.getPropertiesFor(head) : Collections.emptyList()));
     }
 
-    @Extension(ordinal = 1000) @Symbol({"multibranch"})
+    @Extension @Symbol({"multibranch"})
     public static class DescriptorImpl extends MultiBranchProjectDescriptor implements IconSpec {
 
         @NonNull
--- a/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProject.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProject.java
@@ -54,6 +54,7 @@ import org.apache.commons.lang.StringUtils;
 import org.jenkins.ui.icon.Icon;
 import org.jenkins.ui.icon.IconSet;
 import org.jenkins.ui.icon.IconSpec;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProject.java
```diff
@@ -133,7 +134,8 @@ private BranchJobProperty reconstructBranchJobProperty(@NonNull WorkflowJob job,
                 strategy != null ? strategy.getPropertiesFor(head) : Collections.emptyList()));
     }
 
-    @Extension public static class DescriptorImpl extends MultiBranchProjectDescriptor implements IconSpec {
+    @Extension(ordinal = 1000) @Symbol({"multibranch"})
```
jglick, line 137: Why the `ordinal`? Is that somehow related to adding a symbol?

src/test/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProjectTest.java
```diff
@@ -267,4 +271,16 @@ public DescriptorImpl() {
         r.assertLogContains("branch=feature2", b2);
     }
 
+    @Test public void getProjectClassBySymbol(){
```
jglick, line 274: This is not really testing anything not apparent from source code. Unless there is some user-visible effect of the change that can be asserted, may as well just skip the test coverage.

jglick (commented):
Note that the display name currently specifies _Pipeline_ https://github.com/jenkinsci/workflow-multibranch-plugin/blob/e4aa6a85ba848d6759c0a46493d710e74e3272b6/src/main/resources/org/jenkinsci/plugins/workflow/multibranch/Messages.properties#L2 OTOH there is no other seriously maintained multibranch impl (`literate` plugin is long dead, as is `yaml-project`) so probably just `multibranch` is clear enough.

src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProject.java
joseblas: you are right, got confused.

src/test/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProjectTest.java
joseblas: removed.

jglick (approved):

PR #240, opened 2023-03-22
Update parent

jglick:
Supersedes #239. Matches https://github.com/jenkinsci/workflow-cps-plugin/pull/682.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,9 +6,6 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
-    ignore:
-    # Must match transitive version from org.jenkins-ci.plugins:subversion:
-    - dependency-name: org.tmatesoft.svnkit:svnkit-cli
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,6 +6,9 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    ignore:
+    # Must match transitive version from org.jenkins-ci.plugins:subversion:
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -6,6 +6,9 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    ignore:
+    # Must match transitive version from org.jenkins-ci.plugins:subversion:
+    - dependency-name: org.tmatesoft.svnkit:svnkit-cli
```
jtnord, line 11: then why don't we just depend on the version transitively?  are their conflicts - or is there a `depenencyManagement` that is lost due to [MNG-5761](https://issues.apache.org/jira/browse/MNG-5761)?

jtnord (approved):

.github/dependabot.yml
jglick: > then why don't we just depend on the version transitively?

`subversion` depends only on `svnkit` (an old version) and does not pass along the `svnkit-cli` dep.

> are [there] conflicts

Yes, which is why #239 failed.

jglick: We could also pick up https://github.com/jenkinsci/subversion-plugin/pull/276 via https://github.com/jenkinsci/bom/pull/1672 but that would still be a bit too old. Would need to go up to https://github.com/jenkinsci/subversion-plugin/pull/278 via https://github.com/jenkinsci/bom/pull/1789. :eyes:

jtnord (approved):
either way works, I was just wondering if there was a way we can minimize the human interaction needed going forward.

jglick:
Unfortunately it does not appear that way. See note in https://github.com/jenkinsci/plugin-pom/releases/tag/plugin-4.55.

jglick:
(see #238)

PR #287, opened 2024-01-04
Adapt standard Jenkins icons

NotMyFault:
The change proposed adapts the standard Jenkins icons (ionicons) used in core itself. In this case, the icons look most the same, but ionicons support our theme-ability API and are recommended to use.


jglick (approved):
https://github.com/jenkinsci/branch-api-plugin/pull/425#pullrequestreview-1806040819

PR #341, opened 2024-12-19
Allow configuring disableConcurrentBuilds on crated WorkflowJob

madisparn:
Added support for configuring `DisableConcurrentBuildsJobProperty` on the created `WorkflowJob`

### Testing done

Added unit tests
Verified the configuration persistence and labels in UI

### Submitter checklist


jglick:
Is this really necessary? You can already configure the same within the `Jenkinsfile`. Generally we do not offer branch properties unless they add something inherently infeasible to add from the first branch build.

madisparn:
We have a project where we want main branch to run concurrently, but PR builds should not run concurrently, due to resource consumption.

jglick:
Sure, that is routine. You just conditionally include that job property or not, depending on the job/branch name.

PR #577, opened 2026-07-24
Clarify NPE when resolveScm omits targets

Chikzz2001:
When `targets` is omitted from `resolveScm`, Pipeline DataBoundConstructor binding passes `null` and `new ArrayList<>(null)` throws a cryptic NPE (JENKINS-67005 / https://github.com/jenkinsci/github-branch-source-plugin/issues/1380).

Fail fast with a clear message that `targets` is required.

### Testing done

- Added unit test covering null `targets`
- Existing `ResolveScmStepTest` scenarios unchanged

### Submitter checklist
- [x] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/multibranch/ResolveScmStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/multibranch/ResolveScmStepTest.java
@@ -25,19 +25,16 @@
 
 package org.jenkinsci.plugins.workflow.multibranch;
 
+import hudson.model.Result;
 import hudson.model.TopLevelItem;
-import java.util.Collections;
 import jenkins.scm.impl.mock.MockSCMController;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/workflow/multibranch/ResolveScmStepTest.java
```diff
@@ -121,4 +126,16 @@ public void given_nonExistingHeadName_when_invokedWithDefault_then_defaultReturn
             j.buildAndAssertSuccess(job);
         }
     }
+
+    @Test
```
jglick, line 127: More helpfully written as an actual Pipeline script, which would be expected to be a `FAILURE` with a specific log message.

Chikzz2001: Rewrote as a Pipeline script expecting `FAILURE` with that message.

jglick (approved):

PR #582, opened 2026-08-27
`SCMRevisionFetcher` as the one place `SCMRevisionAction` is constructed

jglick:
https://github.com/jenkinsci/branch-api-plugin/issues/728#issuecomment-5347815328


jtnord (approved):
As far as I understand this, it looks OK.

I presume it was tested with both a SCM triggered build of a job and a manual "build now"?

